PR #40, opened 2015-05-13
Enable "gzip" compression

lscheinkman:
In order to lower the user's mobile data usage, we should default to a compressed stream instead of uncompressed one.

If the client added an "Accept-Encoding: gzip" header field to the request then the client is responsible for also decompressing the response stream. Android 'HttpURLConnection' is respecting that contract by only adding automatically compression/decompression to the response when the client does not explicit specify the encoding. See https://android.googlesource.com/platform/libcore/+/jb-release/luni/src/main/java/libcore/net/http/HttpEngine.java

This patch will add "Accept-Encoding: gzip" header field to the request by default, use the compressed stream to calculate the CRC32 and use the uncompressed stream to parse the JSON response.

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/http/HttpRequestFactory.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/http/HttpRequestFactory.java
@@ -1,5 +1,6 @@
 /*
  * Copyright 2011-2015 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2015 Numenta, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License").
  * You may not use this file except in compliance with the License.
@@ -94,17 +95,9 @@ class HttpRequestFactory {
             }
         }
 
-        /*
-         * Amazon DynamoDB sets CRC32 checksum in the header 'x-amz-crc32'. If
-         * compression is turned on, then the checksum is calculated on the
-         * compressed data. On the client side, compression is handled by the
-         * HTTP client. In most cases, the client doesn't have access to
-         * compressed data, so there is no way to compute the checksum of
-         * compressed data unless the client compress it. To get around it,
-         * compression is turned off explicitly.
-         */
+        // Enable "gzip" compression by default
         if (request.getHeaders().get("Accept-Encoding") == null) {
-            request.addHeader("Accept-Encoding", "identity");
+            request.addHeader("Accept-Encoding", "gzip");
         }
 
         Map<String, String> headers = new HashMap<String, String>();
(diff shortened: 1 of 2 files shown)


fosterzhang:
Gzip compression is on by default with HttpUrlConnection on Android. The SDK v2.2.0 used this default behavior. However it caused an issue with DynamoDB because of crc32 checksum as explained in the inline comment. Similar issue was reported for Javascript SDK https://github.com/aws/aws-sdk-js/issues/405. We had to turn it off in v2.2.1 (see [release notes](https://aws.amazon.com/releasenotes/Android/4738890751685832)). This feature has to remain off until we find a solution.

As for your change, I don't think it fixes the crc32 checksum issue. The input stream you get is already uncompressed. There is no need to uncompress it with GZIPInputStream. The client SDK calculates the crc32 checksum based on the uncompressed data while the server (DynamoDB) sets the checksum based on compressed data, and the mismatch will cause CRC32MismatchException.

To test your change on the crc32 checksum issue, you can hit DynamoDB. Make sure the request you make returns data more than 2k, a threshold for gzip compression.

lscheinkman:
@fosterzhang  I've implemented this change specifically for a DynamoDB based application we are developing. It was tested with very large amounts of data and the compression and CRC32 calculation worked like a charm.

`HttpUrlConnection` on Android will only add automatic gzip compression if the HTTP request object does not have `Accept-Encoding: gzip` header. When the `Accept-Encoding: gzip` header is added to the request, `HttpUrlConnection` assumes the client is responsible for decompressing the response and will not decompress the stream. See https://android.googlesource.com/platform/libcore/+/jb-release/luni/src/main/java/libcore/net/http/HttpEngine.java#139 and https://android.googlesource.com/platform/libcore/+/jb-release/luni/src/main/java/libcore/net/http/HttpEngine.java#522
 
This "transparent" compression behavior introduced by android's `libcore` library was responsible for the CRC32 mismatch issue. This change is the solution you are looking for. By making the compression request explicit you can receive the compressed stream from `HttpResponse.getContent()` to calculate the CRC32 and then gunzip the stream to parse the response.

Please try it out and let me know if it works for you.

fosterzhang:
Good to know. I'll test it with Android API level 19+ whose HttpUrlConnection is powered by OkHttp. Hope it will apply too.

danielesegato:
Any ETA on the release of this improvement?

fosterzhang:
@danielesegato: we didn't take the pull request directly, but we applied your idea and modified a little to make it work in general. It is now in v2.2.2 (see [releasenotes](https://aws.amazon.com/releasenotes/Android/4067314458888112)). Thank your very much for your contribution.

PR #80, opened 2015-12-05
Fix issue with Lambda with no input args

### LINKED ISSUE #79: IndexOutOfBoundsException when LambdaFunction has no input arguments
Creating a LambdaFunction with annotation that has no input arguments throws "java.lang.IndexOutOfBoundsException : Invalid array range: 0 to 0" upon invocation.
This is because `invoke` method only checks whether it's null and then tries to access index '0' which does not exists.
This problem happens only with Android version before Lollipop, that could be because of different Java versions.
I've created a pull request which fixes this problem.

boynux:
Fixes #79 
This adds a check to make sure that if Lambda method has no input arguments
it won't throw IndexOutOfBoundsException


fosterzhang:
Thanks for the fix. Pull request is accepted and will be included in the next release.

fosterzhang:
I read the issue #79 you opened. It sounds like this issue may occur on earlier Android version. I'll accept this pull request. Thank you.

PR #105, opened 2016-02-26
Support to get bucket, key for S3 TransferObserver

jhen0409:
The original TransferObserver cannot get bucket name and key, but I need to get bucket name and key to do other jobs.

Also, the [iOS version S3 SDK](http://docs.aws.amazon.com/AWSiOSSDK/latest/Classes/AWSS3TransferUtilityTask.html#//api/name/key) is supported to get bucket name and key.


fosterzhang:
Thanks for the effort. We are evaluating this change.

PR #114, opened 2016-03-17
configurable session pause / resume events + new session creation intervals

oratthumzap:
hi,
we currently use the AWS Mobile Analytics SDK in our multi-activity application.

we made the following modifications to the original SDK:
1. added the ability for developers to configure whether session pause / session resume events will be reported. 
2. added the ability for developers to configure the amount of time (in milliseconds) since the previous session was paused after which a new session will be created.
3. added the ability for developers to configure the amount of time (in milliseconds) between consecutive session resumes that will cause after which a new session will be created.

we implemented these changes in order to reduce the number of events that are reported, and sessions that are created.
the changes are especially important for us since our application is composed of multiple activities that are interchanged between one another, therefore causing numerous session pauses and resumes.

we would appreciate if you will be willing to merge these changes to your SDK so that other developers would be able to enjoy them as well :)


pelegb:
@fosterzhang @wdane 
has this PR been reviewed?
any updates?

Thanks

richabra:
Hi, we are no longer taking feature requests to the Mobile Analytics SDK. All new Mobile Analytics features will be put into the Pinpoint SDK http://docs.aws.amazon.com/pinpoint/latest/developerguide/welcome.html. Please look into integrating and contributing to the Pinpoint SDK https://github.com/aws/aws-sdk-android/tree/master/aws-android-sdk-pinpoint.

PR #127, opened 2016-05-09
Update ApiClientHandler.java

Bill-Niz:
Make constructor public so i can extend ApiClientHandler


fosterzhang:
@Bill-Niz Would you please explain your use case?

Bill-Niz:
Hi,
i'm facing multiple trouble with the ApiclientHandle.
First, in the handle response, when the response is an error, it won't be passe as my Object Error but in String 
`String error = content == null ? "NONE" : IOUtils.toString(content);
            ApiClientException ase = new ApiClientException("Service returned code: "
                    + code + ", error: " + error);`

Second, the qson api crash will trying to convert a date. It only supports **iso8601Format**  aka **alternateIso8601DateFormat** -> `"yyyy-MM-dd'T'HH:mm:ss'Z'"`
But my date is an **iso8601DateFormat** --> `"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"`

karthiksaligrama:
@Bill-Niz please open an issue for the isodate format bug. At this time the plan is not the make the APIClientHandler public

Issue #129, opened 2016-05-10
NoClassDefFoundError on StandardCharsets for API 18 and below


adelpuva:
Hi!
The documentation says that AWS SDK supports Android API 10+, but two classes (https://github.com/aws/aws-sdk-android/search?utf8=✓&q=StandardCharsets)  uses **StandardCharset** that is only available from 19 (http://developer.android.com/reference/java/nio/charset/StandardCharsets.html)

danielesegato:
This:
`StandardCharsets.UTF_8`
should be replaced with:
`Charset.forName("UTF-8")`

fosterzhang:
@adelpuva Thanks for reporting this. We are aware of this issue and have a fix for it (as pointed out by @danielesegato). Stay tuned for the next release.

fosterzhang:
@adelpuva This is fixed in v2.2.16. For detailed changes, please refer to the [change log](https://github.com/aws/aws-sdk-android/blob/master/CHANGELOG.md).

adelpuva:
Thank you!

adelpuva:
In the last SDK (2.3.0) is also user **StandardCharsets** class in _CognitoUser.java_:

`srpAuthResponses.put("PASSWORD_CLAIM_SIGNATURE", new String(Base64.encode(hmac), StandardCharsets.UTF_8));`

PR #130, opened 2016-05-10
Replace StandardCharsets (min SDK 19) with Charset.forName()

danielesegato:
StandardCharsets has been added in Java 1.7 and Android SDK 19 and thus
cause a java.lang.NoClassDefFoundError in Android 18 and below.

Should resolve issue #129


fosterzhang:
@danielesegato Thank you for your contribution. It's better to use `StringUtils.UTF8` as a mean of standardization. A similar fix will be included in the next release.

danielesegato:
Can you point out to your fix commit / commits?

PR #155, opened 2016-06-21
Expose Client Configuration in API Factory and prevent user agent being overridden

fgoncalves:
We do a lot of client identification based on user agent in our api requests. Unfortunately the sdk was overriding this.

This PR prevents this so that if there's a parameter present in the client interface that sets the user agent, then we don't override it. In other words, if an interface is defined as such:

```java
    SomeObject get(@Parameter(name = "User-Agent", location = "header") String userAgent);
```

Then the user agent is not overridden.

The PR also exposes the ``ClientConfiguration`` in the ``ApiClientFactory`` to allow for higher flexibility. In our use case we use it to set the user agent for all clients by default and have the above way to specify it per call.

I'm not sure how to handle versioning here, let me know if you'd like me to bump it.


laurens:
thanks, lgtm +1

fgoncalves:
I'm closing this PR so I can separate the 2 features in 2 prs

laurens:
This PR has been split into two distinct feature changes: #158 and #159

PR #158, opened 2016-06-23
Expose client configuration through api factory

fgoncalves:
In our code base we found it very useful to expose the ``ClientConfiguration`` through the ``ApiClientFactory``. This let's us for example set the user agent for all our calls since we do a lot of client identification based on this header.

We'd like to propose this change. We didn't break the contract and respected the method chaining previously there. One can set the configuration as such:

```java
new ApiClientFactory()
                .clientConfiguration(clientConfiguration)
                .endpoint("some endpoint");
```

Thank you


AutomatedPlayground:
+:b:

laurens:
+1 thanks!

fosterzhang:
Make sense. Thank you.

PR #159, opened 2016-06-23
Preventing User-Agent overriding when the client specifies it

fgoncalves:
We usually do a lot of client identification based on User Agent and therefore we thought that it would be good to be able to specify it and not gett it overridden. To not break the contract and the way the SDK is meant to work what we did is check if a ``Parameter`` annotation, which would set the user agent, is present in the client's definition. If it is, we don't set the user agent in the execution context which will prevent it from being overridden. If such ``Parameter`` annotation is not specified, then the old behaviour is maintained.

In other words, this pr lets the client specify a call like so:

```java
    SomeObject get(@Parameter(name = "User-Agent", location = "header") String userAgent);
```

And the user agent won't be overridden.

Thank you very much for your consideration


fosterzhang:
@fgoncalves This scans the annotation twice. I have a similar fix which should performance slightly better. I'll make it into the next release. Thank you for the contribution.

PR #189, opened 2016-08-17
Null check sessionCredentials before saving

JimmyDahlqvist:
When saving sessionCredentials a NullPointerException
will happen if sessionCredentials is null.
Null check sessionCredentials before saving.

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCachingCredentialsProvider.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCachingCredentialsProvider.java
@@ -452,17 +452,21 @@ public class CognitoCachingCredentialsProvider
             }
         }
 
-        saveCredentials(sessionCredentials,
-                getSessionCredentitalsExpiration().getTime());
+        if (sessionCredentials != null) {
+            saveCredentials(sessionCredentials,
+                    getSessionCredentitalsExpiration().getTime());
+        }
 
         return sessionCredentials;
     }
-    
+
     @Override
     public void refresh() {
         super.refresh();
-        saveCredentials(sessionCredentials,
-                getSessionCredentitalsExpiration().getTime());
+        if (sessionCredentials != null) {
+            saveCredentials(sessionCredentials,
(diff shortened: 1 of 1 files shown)


karthiksaligrama:
@JimmyDahlqvist can you tell me how to reproduce the scenario where the session credentials are null? Ideally if you are using cognito they should never be null

JimmyDahlqvist:
@karthiksaligrama The problem happens if you migrate from unauthenticated to authenticated user, hence user log in with Google Account (federated identity). We in this case calls setLogins with the map with user authentication. This seems to clear the session credentials.

Also when looking at the code the function saveCredentials in https://github.com/aws/aws-sdk-android/blob/23c58741cfbcf8b2f06858b458765fa47abe967e/aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCachingCredentialsProvider.java
already do a null-check. However the null-pointer actually happens due to getSessionCredentitalsExpiration() returning null.
By null-checking before saveCredentials are called this is prevented.

rubinbasha:
Greetings @JimmyDahlqvist .

I just wanted to ask you something about your pull request. We have the same issue as you with the NullPointerException and would be interesting to know how you did solve this and if there was/is/will be a merge happening between your pull and the master branch. 
Also it would be interesting to know what are the consequences of having that null check, does the sdk have issues somewhere else than or does it just not get Push notifications anymore...

Best regards,
Rubin Basha

fernwilter:
Hi @rubinbasha 

We added the null check and we are building the SDK from source in our project.
We have not seen any side effects with it.

Regards
Jimmy

rubinbasha:
Hi @fernwilter thank you for the information, we also did the same thing as you did and are now building the SDK ourselves also. We too are not having any further issue so for the time being we will also keep it with a null check until the SDK handles the issue. I think they are already tackling the issue since they replied to me on this thread [android/issues/247](https://github.com/aws/aws-sdk-android/issues/247)

azsolinsky:
The underlying synchronization issue has been corrected in the AWS SDK for Android 2.3.9 Release. See the release notes for more details: https://aws.amazon.com/releasenotes/1894188094590101

PR #205, opened 2016-10-13
Ignore null request parameters

thearaks:
When building requests, skip parameters which value is null.

This avoid having a number of parameters that contains the String
"null". That was particularly annoying with header parameters.


karthiksaligrama:
This needs to be more smarter. There could be parameters which are non nullable. The SDK needs to account for those

thearaks:
I think that inutil the SDK can't handle non null parameters it's convenient to avoid filling every empty header or pathParam with the String.valueOf(null) value: it is an incorrect behavior.
Anyway, I can explore a way to enforce required parameters to be non null. But I don't know if the current annotations exported from ApiGateway include the needed information...
Thank you for your opinion!

PR #206, opened 2016-10-13
Enforce meaningful User-Agent header values

thearaks:
When setting ExecutionContext user agent, only use current request
user agent header or the value set in the clientConfiguration.

This avoids invalid user agents made up of the interface simple name
or of the request user agent concatenated to the interface simple name.

### DIFF
--- a/aws-android-sdk-apigateway-core/src/main/java/com/amazonaws/mobileconnectors/apigateway/ApiClientHandler.java
+++ b/aws-android-sdk-apigateway-core/src/main/java/com/amazonaws/mobileconnectors/apigateway/ApiClientHandler.java
@@ -106,10 +106,11 @@ class ApiClientHandler implements InvocationHandler {
         final Request<?> request = buildRequest(method, args);
 
         final ExecutionContext context = new ExecutionContext();
-        String userAgent = apiName;
+        String userAgent;
         if (request.getHeaders().containsKey("User-Agent")) {
-            // append it to execution context
-            userAgent += " " + request.getHeaders().get("User-Agent");
+            userAgent = request.getHeaders().get("User-Agent");
+        } else {
+        	userAgent = clientConfiguration.getUserAgent();
         }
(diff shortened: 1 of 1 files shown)


thearaks:
Can please someone review this?
This is a simple, yet effective, change!
Thank you ;)

karthiksaligrama:
The `clientConfiguration.getUserAgent();` is already appended as a part of every http call. So the only part which your PR seems to do is to remove the simple name. I'm not sure why you want to remove the simple name from the User Agent. Its helpful if you have more than 1 api endpoint.

karthiksaligrama:
closing this request

thearaks:
That's because if I set a custom header it's *obvious* that I don't want it to be corrupted by a no-brained if statement.

> I'm not sure why you want to remove the simple name from the User Agent. Its helpful if you have more than 1 api endpoint.

That's because if I set a custom User Agent value there's a valid reason behind my choice and probably I've already added the "which endpoint" information inside my custom header.

Furthermore concatenating the endpoint class name before another user agent value makes the resulting string not very [RFC 2616](https://tools.ietf.org/html/rfc2616#section-14.43) compilant!

```
14.43 User-Agent

   The User-Agent request-header field contains information about the
   user agent originating the request. This is for statistical purposes,
   the tracing of protocol violations, and automated recognition of user
   agents for the sake of tailoring responses to avoid particular user
   agent limitations. User agents SHOULD include this field with
   requests. The field can contain multiple product tokens (section 3.8)
   and comments identifying the agent and any subproducts which form a
   significant part of the user agent. By convention, the product tokens
   are listed in order of their significance for identifying the
   application.
```

Adding a product token that consist only in the endpoint class name (that itself has no much meaning since also that name is auto-generated) to the User Agent causes server-side headaches when trying to pursue the header same purposes (statistical, violation detection, content tailoring) and breaks the order of significance convention!

That said, adding the endpoint class name to the User Agent *should* be optional, at least!

PR #252, opened 2017-02-13
Setted KMS key to new ObjectMetadata instance

ltovar09:
- Added null validation to sseKMSKey property
- Setted sseKMSKey to the resulting ObjectMetadata instance


ltovar09:
The object metadata sseKMSKey was being omitted for the new object, so i included this field
Is this the right way to do it?

karthiksaligrama:
the 2.4.0 has the fix for setting the kms key id.

Issue #329, opened 2017-08-15
ConcurrentModificationException of the HashMap topicListeners in aws-android-sdk-iot


claudiopastorini:
I solved this bug, see the attached pull request.

* What service are you using?
**aws-android-sdk-iot**

* In what version of SDK are you facing the problem?
**2.2.22(but with newer version too, the problem persist)** 

* Is the issue limited to Simulators / Actual Devices?
**No**

* Can your problem be resolved if you bump to a higher version of SDK?
**No, but I resolved using a `ConcurrentHashMap`, please see the attached pull request** 

* Is this problem related to specific Android/OS version?
**No**

* Can you give us steps to reproduce with a minimal, complete, and verifiable example? Please include any specific network conditions that might be required to reproduce the problem.
**The problem is related to the concurrent access to the `topicListeners`. I am facing this problem when a thread subscribes a new topic and, in the meanwhile, another thread is informing about the arrival of a new message.**

* Please include a stacktrace if applicable.

```
MqttException (0) - java.util.ConcurrentModificationException
    at org.eclipse.paho.client.mqttv3.internal.CommsCallback.run(CommsCallback.java:197)
    at java.lang.Thread.run(Thread.java:761)
Caused by: java.util.ConcurrentModificationException
    at java.util.LinkedHashMap$LinkedHashIterator.nextEntry(LinkedHashMap.java:411)
    at java.util.LinkedHashMap$KeyIterator.next(LinkedHashMap.java:422)
    at com.########.android.app.AwsClient$2$4.onMessageArrived(AwsClient.java:248)
    at com.amazonaws.mobileconnectors.iot.AWSIotMqttManager$6.messageArrived(AWSIotMqttManager.java:1233)
    at org.eclipse.paho.client.mqttv3.internal.CommsCallback.deliverMessage(CommsCallback.java:475)
    at org.eclipse.paho.client.mqttv3.internal.CommsCallback.handleMessage(CommsCallback.java:379)
    at org.eclipse.paho.client.mqttv3.internal.CommsCallback.run(CommsCallback.java:183)
	... 1 more
```

kvasukib:
Hi @claudiopastorini,

Thanks for reporting to us. We are looking into your pull request.

Meanwhile, can you provide the steps for reproducing the ConcurrentModificationException?

Thanks

claudiopastorini:
Yeah sure.
[...]

PR #330, opened 2017-08-15
Start to use ConcurrentHashMap in order to avoid ConcurrentModificationException

claudiopastorini:
This solves #329 .


minbi:
This looks good to me. We will merge at next available opportunity

PR #339, opened 2017-09-08
Is CognitoUser logged in method 

### LINKED ISSUE #260: No public method to know if a user is logged in
* What service are you using?
Cognito

* In what version of SDK are you facing the problem?
2.3.8

My issue : I haven't found a way to check if a user is currently logged in via `CognitoUser` class, ie if cached sessions tokens are present on the device.
The `signOut` method allows to clear cached tokens but there doesn't seem to be a public method to know if they are present.
`getCachedSession` would be perfect for this, but it has a protected visibility.

This would be really useful to decide whether to display a Login/Signup action or a Signout action for example.

Prokky:
Resolves #260 feature request

### DIFF
--- a/aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/CognitoUserPoolsSignInProvider.java
+++ b/aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/CognitoUserPoolsSignInProvider.java
@@ -747,4 +747,13 @@ public class CognitoUserPoolsSignInProvider implements SignInProvider {
             return message.substring(0, index);
         }
     }
+    
+    /**
+     * Determines whether the user has an active session or not
+     * 
+     * @return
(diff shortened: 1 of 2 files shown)


kvasukib:
Hi @Prokky,

Thanks for using our SDK and submitting a pull request. We are in the process of reviewing it.

kvasukib:
The logic of checking if a user is logged in is as follows:

```
1. If no token is in cache: return false
2. If token is in cache:
    2.1 check if token is valid by looking into the expiration timestamp:
           if expired return false
           if not expired return true
```

The implementation of the function has to be updated with this. Also, can you resolve the conflicts in the pull request? 

Optionally after checking if the user is logged in, you can write a separate method to check the possibility of the credentials being refreshed as follows:

```
check refresh token, if no refresh token, return false
if refresh token is present, try to refresh, if refresh succeeded, return true
                                                                   if refresh failed, return false
```

Issue #340, opened 2017-09-12
Optimize Kinesis for multiple mixed stream records


themummy:
* What service are you using?
**Kinesis Firehose**
* In what version of SDK are you facing the problem?
**2.6.0**
* Is the issue limited to Simulators / Actual Devices?
**Both**
* Can your problem be resolved if you bump to a higher version of SDK?
**No**
* Is this problem related to specific Android/OS version?
**No**
* Can you give us steps to reproduce with a minimal, complete, and verifiable example? Please include any specific network conditions that might be required to reproduce the problem.
**This is a general issue with the way Kinesis batches and submits records.**
* Please include a stacktrace if applicable.
**N/A**

### Issue
Kinesis Firehose takes a very long time to batch and submit records when records from multiple streams are mixed in the record store.

### Description
Our Android client is submitting data to 14 different Firehose streams. The records are saved into the Kinesis recordStore as soon as they're generated. This result in a fragmented recordStore (i.e. 3 records from _stream1_, followed by 1 record from _stream2_, followed by 2 records from _stream3_).

The way [`submitAll()`](https://github.com/aws/aws-sdk-android/blob/0958a37b4757c41cac40597b973ff417e0b758c4/aws-android-sdk-kinesis/src/main/java/com/amazonaws/mobileconnectors/kinesis/kinesisrecorder/AbstractKinesisRecorder.java#L120) works is that it batches consecutive records belonging to one stream, submits them, then grabs the next batch of consecutive records belonging to another stream. This works well when the records are sorted by stream name. But as one can image, in a fragmented record store, this produces a lot of batches with very few records. This causes `submitAllRecords()` to run for a long time consuming device CPU and power, and very little bandwidth. In our case, it can run for more than an hour.

### Proposed Solution
We can minimize the number of batches being sent by grouping records of the same stream together before sending. We can do so efficiently by iterating through the records in the recordStore and saving them in a batch buffer according to their respective stream. Once the maximum batch size is reached for any of the stream batches, it gets submitted.

We are working on the code changes, and will submit a PR when finished for your review.

kvasukib:
Hi @themummy,

Thanks for reporting to us and using our SDK.

We will wait for the PR and evaluate it.

Thanks

lexlevi17:
Updated: There is now a PR open for this issue.

PR #348, opened 2017-09-25
Optimize Kinesis for multiple mixed stream records (Fix Issue #340)

alevkov:
**Issue**

Kinesis Firehose takes a very long time to batch and submit records when records from multiple streams are mixed in the record store.

**Description**

Our Android client is submitting data to 14 different Firehose streams. The records are saved into the Kinesis recordStore as soon as they're generated. This result in a fragmented recordStore (i.e. 3 records from stream1, followed by 1 record from stream2, followed by 2 records from stream3).

The way submitAll() works is that it batches consecutive records belonging to one stream, submits them, then grabs the next batch of consecutive records belonging to another stream. This works well when the records are sorted by stream name. But as one can image, in a fragmented record store, this produces a lot of batches with very few records. This causes submitAllRecords() to run for a long time consuming device CPU and power, and very little bandwidth. In our case, it can run for more than an hour.

**Offered Solution**

We can minimize the number of batches being sent by grouping records of the same stream together before sending. We can do so efficiently by iterating through the records in the recordStore and saving them in a batch buffer according to their respective stream. Once the maximum batch size is reached for any of the stream batches, it gets submitted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ android:
     - build-tools-24.0.2
     - build-tools-19.1.0
     - android-11
+    - android-15
 sudo: true
 script:
     - sudo rm /etc/mavenrc
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ android:
     - android-15
 sudo: true
 script:
-    - sudo rm /etc/mavenrc
+    - sudo rm -f /etc/mavenrc
     - export MAVEN_OPTS="-Xms512m -Xmx1024m -XX:PermSize=256m -XX:MaxPermSize=512m"
     - mvn clean install -Dmaven.javadoc.skip=true cobertura:cobertura
 
--- a/aws-android-sdk-kinesis/src/main/java/com/amazonaws/mobileconnectors/kinesis/kinesisrecorder/AbstractKinesisRecorder.java
+++ b/aws-android-sdk-kinesis/src/main/java/com/amazonaws/mobileconnectors/kinesis/kinesisrecorder/AbstractKinesisRecorder.java
@@ -25,7 +25,9 @@ import org.apache.commons.logging.LogFactory;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 
 /**
  * An abstract class for Amazon Kinesis recorders. It manages local file store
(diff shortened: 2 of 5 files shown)


lexlevi17:
Hi @minbi,

Thank you for taking your time to review the PR; we appreciate your feedback and are happy to address your concerns.

As you may have gathered from our proposed implementation, we are making a trade-off between an increased occupancy of RAM (i.e. through the usage of a hashtable-backed buffer to store batches) and the reduction of `sendBatch(...)` calls as a function of filling each pending batch in the buffer to max capacity before submission. 

> the code is removing the records before it is ensured the buffer has submitted.

While we do remove read records from the cache after a batch for a given stream name has reached the maxCount/maxSize threshold, all read records that have yet to be submitted remain in the buffer, implying that there should be no unforeseen loss of data. In the instance of `sendBatch(...)` resulting in a recoverable failure, the read records will be kept in the cache, as per the original implementation, which we haven't changed.

> there is no code to submit both streams.

If we reach the end of the file, and the buffer still contains unsubmitted records, the `while` block in `submitAllRecords()` will continue to invoke `nextBatch(...)` until any non-empty batch for a given stream name in the buffer has been submitted. This case, and the general functionality of the buffer, are also covered by our custom unit-tests, which you may refer to in the respective [test class](https://github.com/twosense/aws-sdk-android/blob/master/aws-android-sdk-kinesis/src/test/java/com/amazonaws/mobileconnectors/kinesis/kinesisrecorder/AbstractKinesisRecorderTest.java#L250).

We hope that we have adequately answered your inquiry, and we would be happy to respond to any other concerns with our implementation. Thank you.

-dev team @twosense

lexlevi17:
@minbi @kvasukib The feedback has been addressed.

kvasukib:
Hi @lexlevi17, Thanks for addressing the comment. We are looking into it.

alevkov:
@kvasukib would it be possible to revisit this issue, so that mr stalebot doesn't close it? Thanks.

muellerfr:
Hello @alevkov,

we're still concerned that this change will cause data loss in case of app crashes. We agree with the general idea of this changes. So there are two options here:

1. You can update the PR to not store data in memory, which can cause potential data loss during app crashes.
2. We will update the PR when we have time to do the same.

alevkov:
Hello @muellerfr,

We acknowledge your concerns and we will take care of this issue.

muellerfr:
Thank you, @alevkov.

muellerfr:
@alevkov, do you have an update for this PR?

alevkov:
@muellerfr we will have an update for you soon.

muellerfr:
Hi @alevkov,

I am going to close the PR since we have not seen any additional changes based on our feedback from 10/17. Please open a new PR when your changes are ready.

Thanks,
Frank

PR #387, opened 2017-12-27
Bug in AuthClient.java, method isAuthenticated().

josephrichard7:
Wrong parameter is being sent to LocalDataManager.getCachedSession() method. Second parameter should be the App Client Id instead of App Web Domain.


kvasukib:
Hi @josephrichard7, Thanks for submitting a PR. Can you check the  Travis Build failure and fix it and resubmit the PR?

PR #403, opened 2018-02-01
Decouple Android specific code (cognito)

snodnipper:
ref: https://github.com/aws/aws-sdk-android/issues/384

Android code has been separated to allow both Java and Android projects to
use the same codebase.

This commit is for discussion rather than immediate inclusion - with the underlying premise that the Android code should enhance the Java SDK and not duplicate it.

Note: only tested with (non-committed) gradle config.
      poms updated with eyeballs only!

Summary
* Storage has been abstracted - Android uses SharedPreferences, Java stores in memory.
* Data aggregation has been abstracted - Android stores as before, Java stores JVM details.
* Callback execution has been abstracted - Android uses a handler as before, Java executes on a Thread.
* Demo projects added - one Android example and one Java example

android
console


snodnipper:
I have no problem with the bot closing this issue _but_ the lack of activity largely reflects the state of the whole project.

It is a shame to see an Amazon project like this - it makes developers turn to other solutions.

PR #427, opened 2018-03-22
Fix pubs while reconnecting

sunilpalacherla:
Problem: Publishing silently fails when client is reconnecting and

offline queues are disabled.
Fix: Allow application to handle republishing of failed messages OR
enable offline queues by throwing an exception under these conditions.


minbi:
Hi @sunilpalacherla ,

Is there a reason that you chose to use a thrown exception instead of a callback with error?

PR #438, opened 2018-04-08
fix some errors in RestUtils.java

excellentkf:
The question mark is used when we use the multipartupload function. And it’s meaning to add the question mark in the function, but the condition is impossible, because the buf is not null. Other than, the character '&' should be added into the string, but in the function the variable separator is not used actually.

### DIFF
--- a/aws-android-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/RestUtils.java
+++ b/aws-android-sdk-s3/src/main/java/com/amazonaws/services/s3/internal/RestUtils.java
@@ -171,6 +171,7 @@ public class RestUtils {
                 new String[request.getParameters().size()]);
         Arrays.sort(parameterNames);
         char separator = '?';
+        boolean first = true;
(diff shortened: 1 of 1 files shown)


excellentkf:
hi, @kvasukib , do you mind to help to review it ? thanks

minbi:
Hi @excellentkf ,

The code in this section does not seem to be correct. See sigv4 documentation [here](https://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html)

The canonical request not need to have a question mark for the query string.

Do you have integration tests that demonstrate correctness?

excellentkf:
Hi @minbi , the RestUtils.java is used in  S3Signer.java. It's about sigv2. So, will sigv2 keep update or it  will be abandoned？Thanks for your review.

minbi:
sigv2 has been deprecated in favor of sigv4. Thank you for bringing this to our attention. We can see if this class can be removed to reduce confusion.

PR #456, opened 2018-05-11
Add missing dependency in AWSIotMqttManager

kvasukib:
The PR merge for ConcurrentHashMap removed the import for Map. Added it back in `AWSIotMqttManager`.


minbi:
This is only a repo issue. Maven release and gradle are un-affected.

PR #489, opened 2018-07-04
Cognito: Send secret hash during USER_PASSWORD authentication.

maff91:
### Problem

CognitoUser.getSession initiated with "USER_PASSWORD" authentication type does not send secret hash with the in auth request which leads to: "NotAuthorizedException: Unable to verify secret hash for client".

### Original Issue

https://github.com/aws/aws-sdk-android/issues/484


minbi:
Hi @maff91 ,

Thanks for the pull request. We really appreciate it, but we've already incorporated it into release 2.6.24. Thanks!

PR #495, opened 2018-07-12
Added identity provider as parameter in Auth Builder

faizanmirza:


minbi:
Thanks for the PR. We will do some testing and merge ASAP. This looks good.

shizambles:
Hi,
    When can we expect this PR to be merged? We also have a need to be able to specify the identityprovider (as in the iOS API) so that we can do Facebook authentication. Any way I can help expedite this?

Cheers

PR #532, opened 2018-09-20
AWSIotMqttManager messagequeue

rvp-thunderbuild:
Occasionally the AWSIotMqttManager throws the following exception on the main thread:
`java.lang.NullPointerException: Attempt to read from field 'java.lang.Object java.util.LinkedList$Node.item' on a null object reference
        at java.util.LinkedList.unlink(LinkedList.java:211)
        at java.util.LinkedList.remove(LinkedList.java:526)
        at com.amazonaws.mobileconnectors.iot.AWSIotMqttManager.publishMessagesFromQueue(AWSIotMqttManager.java:1224)`

Which will cause the entire app to crash.
This was reported earlier in [https://github.com/aws/aws-sdk-android/issues/421](url) and never fixed.

 I haven't been able to consistently reproduce this issue because it seems to be timing related where multiple threads are removing items from the message queue. In order to solve this problem I propose the following two fixes:

1) Replace the LinkedList with a ConcurrentLinkedQueue in order to allow better access for multiple threads.
2) Use a looper of a thread different from the main thread, since there is absolutely no need to push queued message from the main thread.

### DIFF
--- a/aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
+++ b/aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
@@ -45,8 +45,7 @@ import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.UnrecoverableKeyException;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.LinkedList;
-import java.util.List;
+import java.util.concurrent.ConcurrentLinkedQueue;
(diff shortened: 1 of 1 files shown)


rvp-thunderbuild:
@kvasukib When will the Looper issue be addressed? And where can I follow this issue?

kvasukib:
@rvp-thunderbuild For the looper change, we need to ensure that the handler thread quits gracefully. Otherwise, there will be a lot of threads lying around without being garbage collected. You are welcome to submit a PR with the additional change of quitting the handler thread appropriately or open a new issue. Thanks.

PR #575, opened 2018-11-13
Allow users to specify the type of connection to use for s3 transfers.

nasdf:
This change allows users to specify what type of connection they'd like to use for S3 transfers.  For example you may only want to use WiFi to upload larger files.

### DIFF
--- a/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferService.java
+++ b/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferService.java
@@ -138,7 +138,7 @@ public class TransferService extends Service {
             if (ConnectivityManager.CONNECTIVITY_ACTION.equals(intent.getAction())) {
                 LOGGER.info("Network connectivity changed detected.");
 
-                final boolean networkConnected = isNetworkConnected();
+                final boolean networkConnected = isNetworkConnected(TransferConnectionType.ANY);
                 LOGGER.info("Network connected: " + networkConnected);
 
                 /**
@@ -149,11 +149,8 @@ public class TransferService extends Service {
                 new Thread(new Runnable() {
                     @Override
                     public void run() {
-                        if (networkConnected) {
-                            checkTransfersOnNetworkReconnect();
-                        } else {
-                            pauseAllForNetwork();
-                        }
+                        pauseAllForNetwork();
+                        checkTransfersOnNetworkReconnect();
                     }
                 }).start();
             }
@@ -162,11 +159,11 @@ public class TransferService extends Service {
         /**
          * Gets the status of network connectivity.
          *
+         * @param connectionType {@link TransferConnectionType}
          * @return true if network is connected, false otherwise.
          */
-        boolean isNetworkConnected() {
-            final NetworkInfo info = connManager.getActiveNetworkInfo();
-            return info != null && info.isConnected();
+        boolean isNetworkConnected(TransferConnectionType connectionType) {
+            return connectionType.isConnected(connManager);
         }
     }
 
@@ -215,7 +212,7 @@ public class TransferService extends Service {
      * and resume them to execution.
      */
     void checkTransfersOnNetworkReconnect() {
-        if (networkInfoReceiver.isNetworkConnected()) {
+        if (networkInfoReceiver.isNetworkConnected(TransferConnectionType.ANY)) {
             loadAndResumeTransfersFromDB(new TransferState[] {TransferState.WAITING_FOR_NETWORK});
         } else {
             LOGGER.error("Network Connect message received but not connected to network.");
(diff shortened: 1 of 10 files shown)


minbi:
Hi @nasdf ,

Thanks for the PR we will start reviewing and update the thread afterward.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -385,15 +385,16 @@ public TransferObserver download(String key, File file) {
      * @param key The key under which the object to download is stored.
      * @param file The file to download the object's data to.
      * @param listener a listener to attach to transfer observer.
+     * @param connectionType The connection type to use for the transfer.
      * @return A TransferObserver used to track download progress and state
      */
     public TransferObserver download(String bucket, String key, File file,
-            TransferListener listener) {
+            TransferListener listener, TransferConnectionType connectionType) {
```
kvasukib: This would be a breaking API change for developers. Instead can you add this option through the TransferUtilityOptions object? See https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtilityOptions.java. You can add a new data member with getter and setter and a default value.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -551,20 +553,21 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata,
      * @param metadata The S3 metadata to associate with this object
      * @param cannedAcl The canned ACL to associate with this object
      * @param listener a listener to attach to transfer observer.
+     * @param connectionType The connection type to use for the transfer.
      * @return A TransferObserver used to track upload progress and state
      */
     public TransferObserver upload(String bucket, String key, File file, ObjectMetadata metadata,
```
kvasukib: This would be a breaking API change for developers. Instead can you add this option through the TransferUtilityOptions object? See https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtilityOptions.java. You can add a new data member with getter and setter and a default value.

kvasukib: Thank you @nasdf for addressing the comments. I will be testing your changes this week and I will have an update based on how the testing goes.

danieldekerlegand:
@kvasukib @cbommas Any updates on this PR? I'm very interested in using this feature in my project. Thanks!

kvasukib:
@danieldekerlegand We are still looking into the changes. I will post an update when the PR is merged and released.

nasdf:
Any updates on this?

muellerfr:
Hi @cbommas,

can you please take a look at this PR?

kvasukib:
@nasdf I have approved the changes. I will be pulling in the changes and write tests and will be including this in one of the future releases. Thank you!

mukeshdadarwal:
@kvasukib Is this feature available now?

mukeshdadarwal:
@cbommas Is this feature available now?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferService.java
```diff
@@ -149,11 +149,8 @@ public void onReceive(Context context, Intent intent) {
                 new Thread(new Runnable() {
                     @Override
                     public void run() {
-                        if (networkConnected) {
```
kvasukib: This needs to be worked on. The main feature of this class is to check for network connectivity and pause when offline and resume when online.

kvasukib:
There are significant changes done in `2.11.0` version of the SDK. The `NetworkInfoReceiver` from `TransferService` is decoupled into a standalone component called `TransferNetworkLossHandler` that can be instantiated without `TransferService`. 

 `TransferNetworkLossHandler` (formerly NetworkInfoReceiver) is a singleton. The connection type is required by  `TransferNetworkLossHandler` to detect the connectivity based on the type. I am looking into the different options on how can the `TransferConnectionType` be available in  `TransferNetworkLossHandler`.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferService.java
nasdf: I moved the network checking logic into the `pauseAllForNetwork `and `checkTransfersOnNetworkReconnect `methods so that each transfer can check if its network type is active or not.  Does that not work with the updated SDK code?

kvasukib: @nasdf Thank you for the information. However my concern is that, when a developer passes the `TransferUtilityOptions` via `TransferUtility` object, it is used for the transfers (upload and download). However the NetworkInfoReceiver / TransferService is not aware of that connection type chosen by the developer and we are checking for ANY.

My suggestion would be to find a way to be consistent with the connection type and use that everywhere in the SDK (TransferUtility, Upload/DownloadTask, TransferService and TransferNetworkLossHandler).

I am working on this suggestion and discussing with other folks in the team. I would like to know your approach for this use-case as well.

bharatUxarmy:
Hey Guys,

I'm waiting for long enough time to get this feature where we can mark Network type - WiFi or mobile data on uploading. If possible for you guys then please let me know when this feature would be available.

Thanks for your time.

kvasukib:
@bharatUxarmy Sorry for the delay. I am working on integrating the changes with the current state of TransferUtility and adding tests. This should be included in the next upcoming release.

kvasukib:
Thank you @nasdf for the PR. This change has been released in 2.11.1 version of the SDK. I'm closing the PR as there are significant changes in the codebase which creates conflicts with the PR branch. Your work is highly appreciated!

PR #579, opened 2018-11-18
Update AwsIotEndpointUtility.java

brandonbai:
input endpoint -> XXXXXX.ats.iot.[region].amazonaws.com.cn

Region.getRegion: throw Cannot create enum from iot value!


brandonbai:
When I use this endpoint like "xxxxx.ats.iot.cn-north-1.amazonaws.com.cn", it throws an exception.
It was later found that the intercepted region was incorrect, it returned "iot" instead of "cn-north-1";

PR #593, opened 2018-11-29
Update CognitoSecretHash.java

NischalManandhar:
While creating CognitoUserPool instance using the awsConfiguration.json file with following description.
"CognitoUserPool": {
    "Default": {
      "PoolId": "xxxxxxxxxx",
      "AppClientId": "xxxxxxxx",
      "AppClientSecret": "",
      "Region": "xxxx"
    }
  }

It will return appClientSecret as "" which is not checked in this file and cause crash in android application.

### DIFF
--- a/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/util/CognitoSecretHash.java
+++ b/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/util/CognitoSecretHash.java
@@ -49,8 +49,9 @@ public final class CognitoSecretHash {
             throw new CognitoParameterInvalidException("client ID cannot be null");
         }
 
-        // Return null as secret hash if clientSecret is null.
-        if (clientSecret == null) {
+        // Return null as secret hash if clientSecret is null or "".
+        // We receive clientSecret as "" if we create cognitoUserPool from awsConfiguration file with "AppClientSecret": "
+        if (TextUtils.isEmpty(clientSecret)) {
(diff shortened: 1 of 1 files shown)


aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/util/CognitoSecretHash.java
```diff
@@ -49,8 +49,9 @@ public static String getSecretHash(String userId, String clientId, String client
             throw new CognitoParameterInvalidException("client ID cannot be null");
         }
 
-        // Return null as secret hash if clientSecret is null.
-        if (clientSecret == null) {
+        // Return null as secret hash if clientSecret is null or "".
+        // We receive clientSecret as "" if we create cognitoUserPool from awsConfiguration file with "AppClientSecret": "
```
palpatim: nit: could you close the last quote? :)

lolsheeplol:
Just throwing in my two cents, but would it make sense to use the already existing `com.amazonaws.util.StringUtils` instead of importing android `TextUtils`?  

There is already a method `public static boolean isBlank(CharSequence cs)` that exists.

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/util/CognitoSecretHash.java
minbi: Changes were made

PR #601, opened 2018-12-06
Prevent HandlerThread leak of AWSIotMqttManager

hirofumi:
Otherwise leak occurs on reconnection as follows.

https://github.com/hirofumi/aws-sdk-android-ht-leak


berniezhao11:
Can confirm it leaks when failed to connect and eventual end up with 500+ threads pending, crashing the app

PR #612, opened 2018-12-12
 Fix reconnection failure upon network disconnect(Issue#584)

desokroshan:
1. Use default value for clean session upon connect and subsequent reconnects if any
2. Update mqttWebSocketURL to use endpoint with HTTP Port


aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -962,7 +955,7 @@ void reconnectToSession() {
 
                 try {
                     final String mqttWebSocketURL = signer
-                            .getSignedUrl(endpoint, clientCredentialsProvider.getCredentials(),
```
cbommas: Can you add a comment on why this being changed from "endpoint" to "endpointWithHttpPort"?

PR #615, opened 2018-12-13
Cognito auth release service

rlatapy-luna:
Allow user to unbind connectionService bound in custom tab before finishing the context used. Avoid this kind of log error: 
`Activity <App Name> has leaked ServiceConnection <ServiceConnection Name>@438030a8 that was originally bound here`


minbi:
Thoughts (not a request for change, yet): This would require an explicit call to unbind the service. I wonder if we could do it as part of the user cancelling or finishing the sign-in event.

rlatapy-luna:
> Thoughts (not a request for change, yet): This would require an explicit call to unbind the service. I wonder if we could do it as part of the user cancelling or finishing the sign-in event.

I don't know how we can be aware of finish sign-in as we are redirect to the app through the deep linking.

PR #641, opened 2019-01-10
[Urgent] [Pinpoint Android SDK] Add missing "withAttributes" to Event payload while generating PutEventsRequest

datkinnguyen:
In EventRecorder.java class, when converting to Event to submit to Pinpoint, the Event payload is missing `attributes` due to lack of copying `attributes` from the AnalyticsEvent.


kvasukib:
This change is released in `2.10.1` version of the SDK. Thank you for pointing this out.

PR #662, opened 2019-01-18
Cognito auth fallback external browser

rlatapy-luna:
Add a fallback on external browser if custom chrome tab are not available.


rlatapy-luna:
Use custom tab from any supported browser (default browser one in priority) instead of forcing chrome package. Fallback to default browser if no browser implement custom tab.

kvasukib:
@rlatapy-luna Sorry for the delayed response. We merged one of your other PRs. Can you rebase your branch with the latest changes from master?

bluetrickpt:
@rlatapy-luna why did you remove the flag "FLAG_ACTIVITY_NEW_TASK" in commit [https://github.com/aws-amplify/aws-sdk-android/pull/662/commits/58ab7bfc6398dcb70694a7e5166e9034d058d462](https://github.com/aws-amplify/aws-sdk-android/pull/662/commits/58ab7bfc6398dcb70694a7e5166e9034d058d462). Without this flag, you cannot call login with the application context, you need the activity context. In my case this essentially means that I have to build the Auth object every-time I need to do a login to pass the new activity context.

rlatapy-luna:
> @rlatapy-luna Sorry for the delayed response. We merged one of your other PRs. Can you rebase your branch with the latest changes from master?

Ok thanks, done

> @rlatapy-luna why did you remove the flag "FLAG_ACTIVITY_NEW_TASK" in commit [58ab7bf](https://github.com/aws-amplify/aws-sdk-android/commit/58ab7bfc6398dcb70694a7e5166e9034d058d462). Without this flag, you cannot call login with the application context, you need the activity context. In my case this essentially means that I have to build the Auth object every-time I need to do a login to pass the new activity context.

Mmm Im not sure why I removed this flag, I had some task issue with firefox custom tab implementation, I need to retest with the flag (on my side I always recreate the Auth object for login/logout...).

bluetrickpt:
> > @rlatapy-luna why did you remove the flag "FLAG_ACTIVITY_NEW_TASK" in commit [58ab7bf](https://github.com/aws-amplify/aws-sdk-android/commit/58ab7bfc6398dcb70694a7e5166e9034d058d462). Without this flag, you cannot call login with the application context, you need the activity context. In my case this essentially means that I have to build the Auth object every-time I need to do a login to pass the new activity context.
> 
> Mmm Im not sure why I removed this flag, I had some task issue with firefox custom tab implementation, I need to retest with the flag (on my side I always recreate the Auth object for login/
>
> edit: Using the FLAG_ACTIVITY_NEW_TASK flag, two task are showing up in the task manager and I cannot restore the task containing the custom tab.. I'm not friendly with task handling, so I prefer to avoid the use of these kind of flag. Maybe we can just add a flag setter like I did for the extra bundle..?

I see. Well, rebuilding the auth for login/logout works. I was just under the impression that since one sends the applicationContext to the builder, it would be more robust if it were to work outside of an activity scope. As it currently is (and correct me if I'm wrong) you have to send the activity as context.

This issue complicates things. The flag setter would work, although it might take people to run into this issue. Maybe this needs further discussion, I'm not experienced enough.

rlatapy-luna:
@kvasukib Is this PR still alive ? Please ask if you want so changes

minbi:
@rlatapy-luna ,

This PR is still alive, we are working on better UI based testing to support multiple browser's CustomTabs. The timeline may be delayed, but this is still under consideration

PR #776, opened 2019-03-11
Pinpoint fix #773

### LINKED ISSUE #773: Submit Events is not working properly
**Describe the bug**
After Investigation I have found that the sdk is not storing the offline data, when device is offline and then if we call the submit event method just after the record event then crash occurs and because of this crash it is unable to store the records in the DB. When I check the awspinpint.db file its empty.

So, I have to add the check network availability whenever I call the submit method. If we don't do this then our data is lost.

**To Reproduce**
Disconnect the device from Internet and then record and submit events then check the db file. Or again connect the device to internet it will not send the previous recorded data.
```
Your code
Same like in the documentation:
https://aws-amplify.github.io/docs/android/analytics#custom-events

public void logEvent() {
   final AnalyticsEvent event =
       pinpointManager.getAnalyticsClient().createEvent("EventName")
           .withAttribute("DemoAttribute1", "DemoAttributeValue1")
           .withAttribute("DemoAttribute2", "DemoAttributeValue2")
           .withMetric("DemoMetric1", Math.random());
   pinpointManager.getAnalyticsClient().recordEvent(event);
   pinpointManager.getAnalyticsClient().submitEvents();
}
```
**Expected behavior**
SDK should handle the submitEvent crash method when device is offline and also record the events in DB properly.

kvasukib:
*Issue #, if available:*

#773 

*Description of changes:*


aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
```diff
@@ -367,43 +380,43 @@ JSONArray getBatchOfEvents(final Cursor cursor, final HashMap<Integer, Integer>
...
-    }
-
-    private void addAllEventIdsToSet(JSONArray eventArray, Set<String> failedEventIds) {
-        for(int i = 0; i < eventArray.length(); i++) {
-            try {
-                failedEventIds.add(eventArray.getJSONObject(i).getString(EVENT_ID));
-            } catch (JSONException e) {
-                log.error("Failed to get event id while processing event item response.", e);
-            }
+            batchIdsAndSizeToDelete.clear();
```
minbi: We should only retry on known exceptions otherwise we may cause in infinite retry.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
```diff
@@ -367,43 +380,43 @@ JSONArray getBatchOfEvents(final Cursor cursor, final HashMap<Integer, Integer>
...
             } else {
                 log.error(
                         String.format("Unable to successfully deliver events to server. " +
-                                "Events will be saved, error is likely recoverable. " +
-                                "Response Status code: %s, Response Error Code: %s",
+                                        "Events will be saved, error is likely recoverable. " +
+                                        "Response Status code: %s, Response Error Code: %s",
                                 amazonServiceException.getStatusCode(), amazonServiceException.getErrorCode()),
                         amazonServiceException);
+                batchIdsAndSizeToDelete.clear();
```
minbi: Main change

TTransmit:
@kvasukib @minbi I can confirm that these changes fix the issue. We are needing to use this as a JAR we have compiled now as this is a major issue for us with business reporting to clients. A lot of our users are in areas with very poor internet connectivity. Is this fix likely to make it into a release soon?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
kvasukib: Currently keeping the events in the local database only when the network is not available. For other exceptions, the exception is logged and the events will be removed from the local database.

PR #779, opened 2019-03-11
Add TLS ALPN support for AWS IoT MQTT connections

desokroshan:
*Issue #, if available:*

*Description of changes:*


aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -172,6 +172,8 @@
      */
     private static final String SDK_VERSION = VersionInfoUtils.getVersion();
 
+    protected static int portNumber = 8883;
```
cbommas: Does this need to be an instance variable? Better to create an internal method that accepts a port number as a parameter.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
```diff
@@ -52,11 +56,23 @@ private AWSIotSslUtility() {
...
      */
+    protected static final String[] ALPN_EXTENSION = {"x-amzn-mqtt-ca"};
+    private static final String TLS_V_1_2 = "TLSv1.2";
+
     public static SSLSocketFactory getSocketFactoryWithKeyStore(KeyStore keyStore)
             throws NoSuchAlgorithmException, UnrecoverableKeyException, KeyStoreException,
             KeyManagementException {
 
-        SSLContext context = SSLContext.getInstance("TLSv1.2");
+        Security.addProvider(new OpenSSLProvider());
```
cbommas: Shouldn't this be done only the ALPN case?

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotTLSSocketFactory.java
```diff
@@ -105,6 +109,9 @@ private Socket enableTLSOnSocket(Socket socket) {
             ((SSLSocket) socket).setEnabledProtocols(new String[] {
                 "TLSv1.2"
             });
+            if (Conscrypt.isConscrypt((SSLSocket) socket)) {
```
cbommas: Only set the ALPN extension if interacting with the server on port 443.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
desokroshan: Removed this variable

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotTLSSocketFactory.java
desokroshan: Discussed offline. We want to single Provider for all the ports and IoT supports ALPN for both 443 and 8883

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
desokroshan: Discussed offline. We want to single Provider for all the cases and IoT supports ALPN for both 443 and 8883

aws-android-sdk-iot/build.gradle
```diff
@@ -5,6 +5,7 @@ dependencies {
 
     api project(":aws-android-sdk-core")
     implementation "org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.1.0"
+    implementation "org.conscrypt:conscrypt-openjdk-uber:1.0.1"
```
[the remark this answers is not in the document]
desokroshan: Yes, the android version does not have the api required to set ALPN flag. Uber version provides convenience in the sense that it contains libraries for all the published platform and provides broader coverage. There are no known vulnerabilities that affect the range of our use-cases.

kvasukib:
This solution changes the behavior for connect using 8883. With this PR, connect using 8883 will be 1) having the ALPN flag 2) using `conscrypt` for creating the socket and all operations associated with the socket. This will be a breaking change for customers using connect using 8883. My suggestion would be to isolate these changes for 443 only.

kvasukib:
@desokroshan The integration test shows conflicts. Can you resolve them so we can see a full pass on integration tests?

desokroshan:
> @desokroshan The integration test shows conflicts. Can you resolve them so we can see a full pass on integration tests?

I have rebased this with the base branch. Integration tests runs after the branch is merged to develop. To run integration test on this branch we will have to change the config file temporarily. Is that what you meant?

aws-android-sdk-iot/pom.xml
```diff
@@ -44,6 +44,11 @@
       <version>1.10.5</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.conscrypt</groupId>
+      <artifactId>conscrypt-android</artifactId>
+      <version>2.0.0</version>
```
minbi:
```
<type>aar</type>
```

Because of all the integration failures that we have seen this change cause in integration tests and  unit tests, has the maven build of this library been tested in an app?

kvasukib:
As discussed add the integration tests for reconnect, disconnect followed by connect for 443.

aws-android-sdk-iot/pom.xml
desokroshan: It has been tested.

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -465,13 +593,18 @@ public void mqttCertificateReconnect() throws Exception {
...
-        mqttManager.connect(ks, new AWSIotMqttClientStatusCallback() {
+        AWSIotMqttClientStatusCallback callback = new AWSIotMqttClientStatusCallback() {
             @Override
             public void onStatusChanged(AWSIotMqttClientStatus status, Throwable throwable) {
                 statuses.add(status);
             }
-        });
+        };
+        // connect to AWS IoT using keystore
+        if (alpnFLag) {
```
kvasukib: alpnFLag -> alpnFlag

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -701,10 +733,10 @@ public void connect(KeyStore keyStore, final AWSIotMqttClientStatusCallback stat
         }
 
         if (endpoint != null) {
-            mqttBrokerURL = String.format("ssl://%s:8883", endpoint);
+            mqttBrokerURL = String.format("ssl://%s:" + portNumber, endpoint);
```
kvasukib: minor: for readability, ```String.format("ssl://%s:%d", endpoint, portNumber);```

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
```diff
@@ -56,7 +84,20 @@ public static SSLSocketFactory getSocketFactoryWithKeyStore(KeyStore keyStore)
...
-        SSLContext context = SSLContext.getInstance("TLSv1.2");
+        SSLContext context;
+
+        if (portNumber == 443) {
+            try {
+                // Attempt to use Conscrypt
+                Security.addProvider(new OpenSSLProvider());
+                context = SSLContext.getInstance(TLS_V_1_2, "Conscrypt");
+            } catch (NoSuchProviderException e) {
+                // Fallback to system SSLContext
```
kvasukib: can you add a log statement saying that the provider does not exist and we are falling back to default provider?

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
```diff
@@ -19,26 +19,54 @@
...
+     * @throws UnrecoverableKeyException when the private key cannot be
+     *             recovered. Ususally a bad keystore password.
+     * @throws KeyStoreException when keystore cannot be created.
+     * @throws KeyManagementException when SSL context cannot be created by key
+     *             manager.
+     */
+    protected static SSLSocketFactory getSocketFactoryWithKeyStore(KeyStore keyStore, int portNum)
+            throws NoSuchAlgorithmException, UnrecoverableKeyException, KeyStoreException,
+            KeyManagementException {
+        portNumber = portNum;
```
kvasukib: You may need a syncrhonization mechanism to protect the assignment of portNumber. For example, when the app has multiple threads T1 and T2; T1 is calling connect and T2 is calling connectUsingALPN. There will be a race condition when the assignment of portNumber is not protected by a lock since portNumber is global state.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
```diff
@@ -15,30 +15,62 @@
...
+     * @param portNum Port number used for connecting to Iot
+     * @return a socket factory for use in creating a secured socket.
+     * @throws NoSuchAlgorithmException when TLS 1.2 is not available.
+     * @throws UnrecoverableKeyException when the private key cannot be
+     *             recovered. Ususally a bad keystore password.
+     * @throws KeyStoreException when keystore cannot be created.
+     * @throws KeyManagementException when SSL context cannot be created by key
+     *             manager.
+     */
+    protected synchronized static SSLSocketFactory getSocketFactoryWithKeyStore(KeyStore keyStore, int portNum)
```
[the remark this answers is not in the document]
desokroshan: The current approach was influenced by reluctance to modify the apparently public method since this wasn't initially planned to be a breaking change but that said I do agree that a utility method should not have variable state that affects other methods. I have retained just `getSocketFactoryWithKeyStore(KeyStore keyStore, int portNum)` method to avoid having to save state.

aws-android-sdk-iot/src/main/res/values/strings.xml
```diff
@@ -1,4 +1,4 @@
 <resources>
-    <string name="app_name">aws-android-sdk-iot-test</string>
+    <string name="app_name">aws-android-sdk-iot</string>
```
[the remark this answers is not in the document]
desokroshan: I think we should keep this for current and future string resources that we might use.

aws-android-sdk-iot/build.gradle
```diff
@@ -1,21 +1,41 @@
-apply plugin: 'java-library'
+apply plugin: 'com.android.library'
 
+android {
+    compileSdkVersion 27
+
+
+
+    defaultConfig {
+        minSdkVersion 21
```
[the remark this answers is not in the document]
desokroshan: Reduced it to 11 to match the minSdkVersion for other android libraries in the SDK

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
```diff
@@ -64,42 +59,15 @@ private AWSIotSslUtility() {
...
-
+            KeyManagementException, NoSuchProviderException {
         SSLContext context;
 
         if (portNumber == 443) {
-            try {
-                // Attempt to use Conscrypt
-                Security.addProvider(new OpenSSLProvider());
-                context = SSLContext.getInstance(TLS_V_1_2, "Conscrypt");
-            } catch (NoSuchProviderException e) {
```
minbi: Why was this catch removed?

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
```diff
@@ -64,42 +59,15 @@ private AWSIotSslUtility() {
...
-                // Fallback to system SSLContext
-                LOGGER.debug("Conscrypt provider not found. Falling back to the default security provider.");
-                context = SSLContext.getInstance(TLS_V_1_2);
-            }
+            // Use Conscrypt as security provider
+            Security.addProvider(new OpenSSLProvider());
```
minbi: What is the behavior of calling `addProvider` multiple times? Does it need to be avoided?

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotTLSSocketFactory.java
```diff
@@ -110,7 +109,7 @@ private Socket enableTLSOnSocket(Socket socket) {
             ((SSLSocket) socket).setEnabledProtocols(new String[] {
                 "TLSv1.2"
             });
-            if (portNumber == 443 && Conscrypt.isConscrypt((SSLSocket) socket)) {
```
minbi: Why was this port number check removed?

aws-android-sdk-iot/src/main/res/values/strings.xml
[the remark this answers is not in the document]
minbi: A strings.xml can be added when we need them. We are not defining anything in this file that is useful.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
desokroshan: It was instead added to the method signature since the default provider will not work with on port 443 as service mandates alpn flag for 443

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotSslUtility.java
desokroshan: No, it must be called each time socketFactory needs to be created

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotTLSSocketFactory.java
desokroshan: Because port 443 nows corresponds to socket having Conscrypt as security provider and vice-versa rendering the check redundant

aws-android-sdk-iot/src/main/res/values/strings.xml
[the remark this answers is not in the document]
desokroshan: It has app_name which is used to get appropriate iot-specific configuration from test configuration file. This is the pattern followed by all other integration tests and there is no reason for this to be an exception.

kvasukib:
@desokroshan Can you fix the conflict in `config.yml` before merging?

PR #782, opened 2019-03-11
build android sdk with android-10

sunchunqiang:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -188,6 +188,39 @@ jobs:
           key: Libs-{{ .Revision }}
           paths:
             - lib
+  build_api10:
+    working_directory: ~/code
+    docker:
+      - image: circleci/android:api-27-alpha
+    environment:
+      JVM_OPTS: -Xmx3200m
+    steps:
+      - checkout
+      - generate_gradle_wrapper
+      - run:
+          name: download android-10
+          command:
+            sudo yes | sdkmanager "platforms;android-10"
+      - run:
+          name: install python3-pip
+          command: |
+            sudo apt-get update
+            sudo apt-get -y install python3-pip            
+      - run:
+          name: install json parser
+          command: sudo pip3 install demjson           
+      - run:
+          name: update code for api10
+          command: |
+            python3 CircleciScripts/replace_android10.py "$(pwd)"
+            rm aws-android-sdk-core/src/main/java/com/amazonaws/internal/keyvaluestore/KeyProvider18.java
+            rm aws-android-sdk-core/src/main/java/com/amazonaws/internal/keyvaluestore/KeyProvider23.java
+            difference=$(git diff)
+            echo "$difference"
+      - run:
+          name: build the whole project
+          command: |
+            bash gradlew :aws-android-sdk-core:build  -x test
 
   unittest:
     working_directory: ~/code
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -188,6 +188,23 @@ jobs:
...
+    working_directory: ~/code
+    docker:
+      - image: circleci/android:api-27-alpha
+    environment:
+      JVM_OPTS: -Xmx3200m
+    steps:
+      - checkout
+      - generate_gradle_wrapper
+      - run:
+          name: downloand android-10
```
minbi: download

.circleci/config.yml
```diff
@@ -188,6 +188,23 @@ jobs:
...
+          command:
+            sudo yes | sdkmanager "platforms;android-23"
```
minbi: android-10

.circleci/config.yml
```diff
@@ -188,6 +188,23 @@ jobs:
...
+      - run:
+          name: build the whole project
+          command: |
+            bash gradlew build -x test
```
minbi: this needs to build just core

.circleci/config.yml
```diff
@@ -603,6 +620,7 @@ workflows:
   build_test:
     jobs:
       - build
+      - build_api10
```
minbi: this should be added to release_sdk as well

CircleciScripts/replace_android10.py
```diff
@@ -0,0 +1,24 @@
+from utils import replacefiles
+import demjson
+import sys
+
+root = sys.argv[1]
+
+replaces = [
+    {
+        "match" : 'android-23',
```
minbi: there needs to be a way to remove the relevant code that uses api 23 and 18. This code is safe-guarded during runtime, but this ensure the rest of the code adheres to api level 10. Check with @kvasukib  which code parts should be removed

.circleci/config.yml
sunchunqiang: addressed

.circleci/config.yml
sunchunqiang: addressed

CircleciScripts/replace_android10.py
sunchunqiang: addressed

.circleci/config.yml
```diff
@@ -1216,6 +1256,7 @@ workflows:
       - release_s3:
           requires:
             - build
+            - build_api10
```
minbi: this should be required by release_maven, not s3

CircleciScripts/replace_android10.py
```diff
@@ -0,0 +1,45 @@
...
+replacefiles(root, replaces)
+
+
+
+
+AWSKeyValueStoreFile = os.path.join(root,"aws-android-sdk-core/src/main/java/com/amazonaws/internal/keyvaluestore/AWSKeyValueStore.java")
+newcontent = ""
+with open(AWSKeyValueStoreFile, 'r') as myfile:
+    content = myfile.read() 
+    pattern = r'if\s*\(apiLevel >= ANDROID_API_LEVEL_23\)\s*\{[^\}]+\}\s*else if\s*\(apiLevel >= ANDROID_API_LEVEL_18\)\s*\{[^\}]+\}'
```
minbi: @kvasukib had added some annotations something along the lines of `@api23start` and `@api23end` would that be a better way to replace?

PR #822, opened 2019-03-27
models-update-iot-2019-03-27-23-41

sunchunqiang:
models-update-iot-2019-03-27-23-41

### DIFF
--- a/aws-android-sdk-iot/src/main/java/com/amazonaws/services/iot/model/CreateScheduledAuditRequest.java
+++ b/aws-android-sdk-iot/src/main/java/com/amazonaws/services/iot/model/CreateScheduledAuditRequest.java
@@ -77,6 +77,13 @@ public class CreateScheduledAuditRequest extends AmazonWebServiceRequest impleme
 
     /**
      * <p>
+     * Metadata which can be used to manage the scheduled audit.
+     * </p>
+     */
+    private java.util.List<Tag> tags;
+
+    /**
+     * <p>
      * The name you want to give to the scheduled audit. (Max. 128 chars)
      * </p>
      * <p>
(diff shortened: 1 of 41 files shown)


aws-android-sdk-iot/src/main/java/com/amazonaws/services/iot/model/CreateScheduledAuditRequest.java
```diff
@@ -75,6 +75,13 @@
      */
```
kvasukib: Just double check if its just moving the code to a different place in the same file.

aws-android-sdk-iot/src/main/java/com/amazonaws/services/iot/model/CreateScheduledAuditRequest.java
```diff
@@ -696,9 +696,9 @@ public int hashCode() {
         hashCode = prime * hashCode + ((getDayOfWeek() == null) ? 0 : getDayOfWeek().hashCode());
         hashCode = prime * hashCode
                 + ((getTargetCheckNames() == null) ? 0 : getTargetCheckNames().hashCode());
+        hashCode = prime * hashCode + ((getTags() == null) ? 0 : getTags().hashCode());
```
minbi: This will change the hashCode calculation

aws-android-sdk-iot/src/main/java/com/amazonaws/services/iot/model/CreateScheduledAuditRequest.java
minbi: Looks like it was a code move.

PR #825, opened 2019-03-28
models-update-pinpoint-2019-03-27-23-41

sunchunqiang:
models-update-pinpoint-2019-03-27-23-41


aws-android-sdk-pinpoint/src/main/java/com/amazonaws/services/pinpoint/model/transform/PutEventsResultJsonUnmarshaller.java
```diff
@@ -1,5 +1,5 @@
 /*
```
kvasukib: This file is hand modified because the codegen couldn't handle unmarshalling of an object that has more than one level of nesting. We need to mark this class as custom in the config so codegen doesn't override the changes in this class.

minbi:
Closing for an update PR.

PR #874, opened 2019-04-05
Fix isAuthenticated in CognitoAuth caused by scopes unmarshalling

### LINKED ISSUE #813: Cognito Auth isAuthenticated always returning false
**Describe the bug**
Function isAuthenticated() from cognitoauth returns always false even after successful login.

**To Reproduce**
Download and configure the cognitoauth sample for android https://github.com/awslabs/aws-sdk-android-samples. 
Add the following line to the onSuccess of the AuthHandler:
```
Log.d(TAG, "Is user signed = " + auth.isAuthenticated());
```
Authenticate and see the logs. _false_ is returned even after a successful login.

**Which AWS service(s) are affected?**
cognitoauth

**Expected behavior**
isAuthenticated() should return true after a successful login.

**Environment Information (please complete the following information):**
 - AWS Cognito Auth SDK Version: 2.12.+
 - Device: Simulator, but should occur in any
 - Android Version: 9.0 (haven't tried in others, but suspect that the same will occur)

**Additional context**
This only occurs in the latest version of cognito auth sdk (2.12.+). In version 2.11.+ isAuthenticated() works as expected.
Similar issue to https://github.com/aws-amplify/aws-sdk-android/issues/508 but that one was probably fixed long ago.

minbi:
Fixes #813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@
 * **AWS Mobile Client**
   * Fixed a bug that caused repetitive sign-in using the drop-in UI to the same provider to not federate the correct credentials. See [issue #809](https://github.com/aws-amplify/aws-sdk-android/issues/809)
 
+* **Amazon Cognito Auth**
+  * Fixed a bug that cause `isAuthenticated()` to return false after sign-in when no scopes were requested. See [issue #813](https://github.com/aws-amplify/aws-sdk-android/issues/813)
+
 * **Amazon Pinpoint**
   * Fix a bug where the image that is part of a push notification is persisted in the subsequent notifications.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
 * **AWS Mobile Client**
   * Fixed a bug that caused repetitive sign-in using the drop-in UI to the same provider to not federate the correct credentials. See [issue #809](https://github.com/aws-amplify/aws-sdk-android/issues/809)
 
+* **Amazon Cognito Auth**
+  * Fixed a bug that cause `isAuthenticated()` to return false after sign-in when no scopes were requested. See [issue #813](https://github.com/aws-amplify/aws-sdk-android/issues/813)
```
rohandubal:
```suggestion
  * Fixed a bug that caused `isAuthenticated()` to return false after sign-in when no scopes were requested. See [issue #813](https://github.com/aws-amplify/aws-sdk-android/issues/813)
```

PR #876, opened 2019-04-05
[skip test] run git push for merge develop to master

rohandubal:
*Description of changes:*

Master should not have any changes when we checkout, so doing a git push.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -891,6 +891,8 @@ jobs:
             git status
             git clean -fd
             git status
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -891,6 +891,8 @@ jobs:
             git status
             git clean -fd
             git status
+            git push https://${GITHUB_TOKEN}@github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}.git
```
minbi: This only pushed the same branch onto itself, you're missing a merge

rohandubal: This is intentional. I believe CircleCI is auto-merging the master branch

Issue #231, opened 2016-12-16
Fatal Exception: com.amazonaws.AmazonClientException


alirezaafkar:
`
Fatal Exception: com.amazonaws.AmazonClientException: Unable to execute HTTP request: Unable to resolve host "cognito-identity.us-east-1.amazonaws.com": No address associated with hostname
       at com.amazonaws.http.AmazonHttpClient.executeHelper(Unknown Source:424)
       at com.amazonaws.http.AmazonHttpClient.execute(Unknown Source:199)
       at com.amazonaws.services.cognitoidentity.AmazonCognitoIdentityClient.invoke(Unknown Source:558)
       at com.amazonaws.services.cognitoidentity.AmazonCognitoIdentityClient.getCredentialsForIdentity(Unknown Source:388)
       at com.amazonaws.auth.CognitoCredentialsProvider.populateCredentialsWithCognito(Unknown Source:1651)
       at com.amazonaws.auth.CognitoCredentialsProvider.getCredentials(Unknown Source:371)
       at com.amazonaws.auth.CognitoCachingCredentialsProvider.getCredentials(Unknown Source:441)
       at com.amazonaws.mobile.user.IdentityManager$AWSCredentialsProviderHolder.getCredentials(Unknown Source:61)
       at com.amazonaws.services.sns.AmazonSNSClient.invoke(Unknown Source:2818)
       at com.amazonaws.services.sns.AmazonSNSClient.setEndpointAttributes(Unknown Source:1455)
       at com.amazonaws.mobile.push.PushManager.setSNSEndpointEnabled(Unknown Source:289)
       at com.amazonaws.mobile.push.PushManager.setPushEnabled(Unknown Source:301)
`

karthiksaligrama:
Steps to reproduce?

alirezaafkar:
pushManager.registerDevice();
                if (pushManager.isRegistered()) {
                    pushManager.setPushEnabled(true);
                    for (Map.Entry<String, SnsTopic> entry : pushManager.getTopics().entrySet()) {
                        pushManager.subscribeToTopic(entry.getValue());
                    }
                }

karthiksaligrama:
I just downloaded and tried a mobile hub sample, and i'm unable to reproduce the problem. Can you give me more information on how to reproduce this?

alirezaafkar:
I can't reproduce it too! Happened for users and I've logs from Crashlyrics.

karthiksaligrama:
[...]

PR #884, opened 2019-04-08
Update CognitoUser.java

hakansuluoglu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-android-sdk-cognitoauth/src/androidTest/java/com/amazonaws/mobileconnectors/cognitoauth/LocalDataManagerIntegrationTest.java
+++ b/aws-android-sdk-cognitoauth/src/androidTest/java/com/amazonaws/mobileconnectors/cognitoauth/LocalDataManagerIntegrationTest.java
@@ -214,7 +214,52 @@ public class LocalDataManagerIntegrationTest extends CognitoAuthIntegrationTestB
                     setFromString("email,profile,openid"));
 
             printSharedPreferencesKeys();
-            verifySharedPreferencesForCachedState(auth.awsKeyValueStore);
+            verifySharedPreferencesForCachedState(auth.awsKeyValueStore, "proofKey", "email,profile,openid");
+        } catch (Exception ex) {
+            ex.printStackTrace();
+            assertTrue("Error occurred" + ex, false);
+        }
+    }
+
+    @Test
+    public void testCognitoAuthCachedStateForNoScopes() {
+        printSharedPreferencesKeys();
+
+        try {
+            Auth.Builder builder = new Auth.Builder();
+            Auth auth = builder.setAppClientId(getPackageConfigure().getString("AppClientId"))
+                    .setAppCognitoWebDomain(getPackageConfigure().getString("WebDomain"))
+                    .setSignInRedirect(getPackageConfigure().getString("SignInRedirectURI"))
+                    .setSignOutRedirect(getPackageConfigure().getString("SignOutRedirectURI"))
+                    .setApplicationContext(InstrumentationRegistry.getTargetContext())
+                    .setAdvancedSecurityDataCollection(true)
+                    .setScopes(new HashSet<String>())
+                    .setAuthHandler(new AuthHandler() {
+                        @Override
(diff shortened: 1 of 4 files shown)


aws-android-sdk-cognitoauth/src/androidTest/java/com/amazonaws/mobileconnectors/cognitoauth/LocalDataManagerIntegrationTest.java
```diff
@@ -214,7 +214,52 @@ public void onFailure(Exception e) {
...
-            verifySharedPreferencesForCachedState(auth.awsKeyValueStore);
+            verifySharedPreferencesForCachedState(auth.awsKeyValueStore, "proofKey", "email,profile,openid");
+        } catch (Exception ex) {
+            ex.printStackTrace();
+            assertTrue("Error occurred" + ex, false);
+        }
+    }
+
+    @Test
+    public void testCognitoAuthCachedStateForNoScopes() {
```
kvasukib: can you rename this to `testCognitoAuthCachedStateForEmptyScopes`?

minbi: The test name is from our code base.

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
```diff
@@ -952,6 +952,7 @@ protected CognitoUserSession getCachedSession() {
                     clearCachedTokens();
                     throw new CognitoNotAuthorizedException("User is not authenticated", nae);
                 } catch (final Exception e) {
+                    clearCachedTokens();
```
minbi: What scenario are you hitting that requires the tokens to be cleared?

hakansuluoglu: I can't find the cause, but user information is not deleted from cache after logging out. (I checked the shared preferences file and checked). The same user cannot login again. The problem is correcting after cache is deleted. We are using user pool without using the Identity pool. I didn't find out how to delete Cache.

hakansuluoglu: Class: Cognito User /  line :2323 

cognitoIdentityProviderClient .initiateAuth(initiateAuthRequest)
returns
'com.amazonaws.services.cognitoidentityprovider.model.UserNotFoundException' exception.

Screen Shot 2019-04-09 at 10 27 36

hakansuluoglu: https://github.com/aws-amplify/aws-sdk-android/issues/231

minbi: Can you change your edit by creating another catch statement specifically for UserNotFoundException? This is because network not being available also manifests in an exception and we want to keep the session in that case.

Can you share which keys are not being deleted?

hakansuluoglu: I think I've found that why user information remains in shared preferences.

In app **AndroidManifest.xml**
`android:allowBackup="true"`

If user deletes the app while logged in, user's information remains in shared preferences file (CognitoIdendityProviderCache). If user reinstall  after Session has dropped, it cannot log in with this information.  Still I don't know why throw UserNotFoundException. 

I changed manifest file; 

`android:allowBackup="false"`
and
`tools:replace="android:allowBackup"`

Issue #964, opened 2019-05-17
AWSMobileClient not working on Android 4.3


luanalbineli:
**Describe the bug**
When we tries to run the AWSMobileClient on 4.2 and 4.3 Android devices, a `NoSuchMethodError` raises:

`java.lang.NoSuchMethodError: android.security.KeyPairGeneratorSpec$Builder.setKeySize
        at com.amazonaws.internal.keyvaluestore.KeyProvider18.a(KeyProvider18.java:154)
        at com.amazonaws.internal.keyvaluestore.KeyProvider18.a(KeyProvider18.java:72)
        at com.amazonaws.internal.keyvaluestore.AWSKeyValueStore.a(AWSKeyValueStore.java:138)
        at com.amazonaws.internal.keyvaluestore.AWSKeyValueStore.<init>(AWSKeyValueStore.java:108)
        at com.amazonaws.mobile.client.AWSMobileClientStore.<init>(AWSMobileClient.java:3462)
        at com.amazonaws.mobile.client.AWSMobileClient$2.run(AWSMobileClient.java:477)
        at com.amazonaws.mobile.client.internal.InternalCallback$1.run(InternalCallback.java:101)
        at java.lang.Thread.run(Thread.java:841)`

**To Reproduce**
Just tries to run the AWSMobileClient on specific 4.3 devices.

**Expected behavior**
Run on Android 4.3, because accordingly with the documentation, the minSdkVersion is 15.

**Environment Information (please complete the following information):**
 - AWS Android SDK Version: 2.13.4
 - Device: I've tested in Samsung Galaxy S3 (4.3), Sony Xperia Z (4.3)
 - Android Version: 4.3
 - Specific to simulators: No

rodrigoarias:
I'm having the same issue.

rodrigoarias:
That's not the only one.
I'm getting this crashes, mostly in Android 4, some on Android 5:

**KeyProvider18.java line 154**
`Fatal Exception: java.lang.NoSuchMethodError: android.security.KeyPairGeneratorSpec$Builder.setKeySize
       at com.amazonaws.internal.keyvaluestore.KeyProvider18.initializeRSAKeyFromKeyStore(KeyProvider18.java:154)
       at com.amazonaws.internal.keyvaluestore.KeyProvider18.getKey(KeyProvider18.java:72)
[...]

PR #967, opened 2019-05-19
[Core] Remove call to setkeySize in KeyProvider18

kvasukib:
*Issue #, if available:*

#964 

*Description of changes:*

`setKeySize` is not available in API Level 18. The default key size used by RSA in `KeyPairGeneratorSpec` is 2048 when not specified. This change will use the default of 2048 in all API levels (18-22).

Added build automation for verifying if the fix works for API 18-22.

### DIFF
--- a/aws-android-sdk-core/build.gradle
+++ b/aws-android-sdk-core/build.gradle
@@ -26,12 +26,14 @@ dependencies {
 // The following command (is idempotent)
 // removes the libs folder, creates libs folder and
 // creates a symbolic link to android-23.jar.
-task myTask(type:Exec) {
+task getApi23(type:Exec) {
+    print 'Creating a symlink for Android API Level 23.'
     workingDir '.'
     commandLine 'sh', '-c', "rm -rf libs && mkdir libs && ln -s $System.env.ANDROID_HOME/platforms/android-23/android.jar libs/android-23.jar"
 }
 
-compileJava.dependsOn(myTask)
+
+compileJava.dependsOn(getApi23)
 
 sourceCompatibility = "1.7"
 targetCompatibility = "1.7"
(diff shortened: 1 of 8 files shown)


aws-android-sdk-core/src/main/java/com/amazonaws/internal/keyvaluestore/KeyProvider18.java
```diff
@@ -151,7 +151,6 @@ private void initializeRSAKeyFromKeyStore(Context context,
                     .setSerialNumber(BigInteger.TEN)
                     .setStartDate(start.getTime())
                     .setEndDate(end.getTime())
-                    .setKeySize(RSA_KEY_SIZE)
```
minbi: The RSA_KEY_SIZE constant definition should also be removed. Potentially update the javadoc regarding the key size and how the default is used. Did we rule out splitting KeyProvider18 into 18 and 19?

kvasukib: Removed the constant and updated the Javadoc. Since the default key size is 2048 bits, I recommend to not invoke the method and keep the implementation the same for 18-22.

SalusaSecondus: Agreed. The default size will not be lowered in the future.

aws-android-sdk-core/build.gradle
```diff
@@ -26,12 +26,18 @@ dependencies {
...
     commandLine 'sh', '-c', "rm -rf libs && mkdir libs && ln -s $System.env.ANDROID_HOME/platforms/android-23/android.jar libs/android-23.jar"
 }
 
-compileJava.dependsOn(myTask)
+task getApi18(type:Exec) {
+    workingDir '.'
+    commandLine 'sh', '-c', "rm -rf libs && mkdir libs && ln -s $System.env.ANDROID_HOME/platforms/android-18/android.jar libs/android-18.jar"
+}
+
+compileJava.dependsOn(getApi23)
```
minbi: Depending on both would defeat the purpose of the separate builds.

kvasukib: Updated.

PR #971, opened 2019-05-21
Remove trailing zeros before putting frame into SDK

zhiyua-git:
*Issue #, if available:*
581, 909
*Description of changes:*
The fix is removing the trailing zeros from encoder output which shows up in some android devices. The multiple trailing zeros were causing 0x3200000d for Kinesis Video.

### DIFF
--- a/aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/producer/KinesisVideoFrame.java
+++ b/aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/producer/KinesisVideoFrame.java
@@ -69,6 +69,9 @@ public class KinesisVideoFrame {
         mPresentationTs = presentationTs;
         mDuration = duration;
         mData = requireNonNull(data);
+        // In some devices encoder would generate frames with more than 3 trailing zeros
+        // which is not allowed by AnnexB specification
+        removeTrailingZeros();
     }
 
     public int getIndex() {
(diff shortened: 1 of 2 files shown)


aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/producer/KinesisVideoFrame.java
```diff
@@ -69,6 +69,9 @@ public KinesisVideoFrame(int index, int flags, long decodingTs, long presentatio
         mPresentationTs = presentationTs;
         mDuration = duration;
         mData = requireNonNull(data);
+        // In some devices encoder would generate frames with more than 3 trailing zeros
+        // which is not allowed by AnnexB specification
+        removeTrailingZeros();
```
minbi: @MushMal I'm not familiar with the AnnexB spec. Could trimming all the zeroes be an issue? Should there be a safety-maximum number of bytes trimmed?

Everything else looks to be in order. Approved on my side.

MushMal: @minbi the important termination condition is in the code - not to overrun the beginning. The rest of the validations will be performed in the lower layers. So, if we end up with no bytes, we will eventually get an error from the bottom layer. Similarly, if we end up with semantically incorrect Annex-B format, it will be caught in the native layer. I think adding extra validation for Annex-B format here is not needed.

Issue #1065, opened 2019-07-06
Error execute createMediaSource using sdk version 2.13.+


addodelgrossi:
**Describe the bug**
When you run the `kinesisVideoClient.createMediaSource` method, the error is returned:

```
E / KinesisVideoAndroidClient: 2019-07-06T15: 12Z T2202: EXCEPTION: ResourceNotFoundException: Service call failed .: The requested stream is not found or not active. (Service: AWSKinesisVideo; Status Code: 404; Error Code: ResourceNotFoundException; Request ID: ba0da9da-faea-461f-919a-354ec7c48a42)
```

The problem occurs when using sdk **com.amazonaws: aws-android-sdk-kinesisvideo: 2.13.** +, If you use the **com.amazonaws: aws-android-sdk-kinesisvideo: 2.11.**  The problem does not occur.

**To Reproduce**
You can reproduce the problem using the AWS AmazonKinesisVideoDemoApp sample project.

```
            mKinesisVideoClient = KinesisVideoAndroidClientFactory.createKinesisVideoClient(
                    getActivity(),
                    KinesisVideoDemoApp.KINESIS_VIDEO_REGION,
                    KinesisVideoDemoApp.getCredentialsProvider());

            mCameraMediaSource = (AndroidCameraMediaSource) mKinesisVideoClient
                    .createMediaSource(mStreamName, mConfiguration);
```

**Which AWS service(s) are affected?**
Kinesis Video Stream

**Environment Information (please complete the following information):**
 - AWS Android SDK Version: 2.13.+
 - Device: Pixel XL, Simulator
 - Android Version: API 28
 - Specific to simulators: No

kvasukib:
@zhiyua-git Can you have a look at this? Thanks.

addodelgrossi:
@zhiyua-git some idea ? Unfortunately only working with version 11, when updating occurs the error above. Thanks.

palpatim:
The fix for this has been released on https://github.com/aws-amplify/aws-sdk-android/releases/tag/release_v2.14.1

Please let us know if you have any questions.

zhiyua-git:
Hi @palpatim , Can you also update the version to be 2.14+? Thanks.
https://github.com/awslabs/aws-sdk-android-samples/blob/master/AmazonKinesisVideoDemoApp/build.gradle#L50

palpatim:
[...]

PR #1084, opened 2019-07-23
Fix Tags Field in CreateStreamRequest Caused ValidationException

zhiyua-git:
*Issue #, if available:*
https://github.com/aws-amplify/aws-sdk-android/issues/1065

*Description of changes:*
Set the tags value to null so Kinesis Video will not throw ValidationException on this.


aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/mobileconnectors/kinesisvideo/service/KinesisVideoAndroidServiceClient.java
```diff
@@ -192,7 +192,8 @@ public String createStream(@NonNull final String streamName,
                 .withDeviceName(deviceName)
                 .withMediaType(contentType)
                 .withKmsKeyId(StringUtils.isBlank(kmsKeyId) ? null : kmsKeyId)
-                .withDataRetentionInHours((int) retentionPeriodInHours);
+                .withDataRetentionInHours((int) retentionPeriodInHours)
+                .withTags(null);
```
kvasukib: Will there be a case where tags can be passed? Can the developer override this?

zhiyua-git: No. It is not expected to be passed on creation now. Passing the tags on creation is complex change also related to SDK state machine change. This feature might be support later.

kvasukib: Can you add a comment in the code to explain why tags are passed as null?

kvasukib: Also, please add a statement in CHANGELOG.md file as a bug fix.

zhiyua-git: Updated.

Issue #1085, opened 2019-07-23
Add public method to determine if CognitoUser is signed in with valid session in Android SDK


spencerelliott:
**Which AWS Services is the feature request for?**
Android Cognito User Pools SDK

**Is your feature request related to a problem? Please describe.**
It is not feasible to use `AWSMobileClient` for my current project, so I am not able to use the workaround provided in #260. I need to be able to see if a user has signed in and has valid credentials for my logic.

**Describe the solution you'd like**
Either make `CognitoUser.getCachedSession` public, or provide a `CognitoUser.isSignedIn` method similar to the iOS Cognito Identity Provider SDK.

gpanshu:
AWSMobileClient is now deprecated in V2 of Amplify. I would recommend that you move to Amplify V2 SDK for Android and use Amplify.Auth.SignIn and fetchAuthSession to determine the current session.

PR #1087, opened 2019-07-24
Make getCachedSession public

desokroshan:
*Issue #, if available:*
https://github.com/aws-amplify/aws-sdk-android/issues/1085

*Description of changes:*

### DIFF
--- a/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
+++ b/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
@@ -925,7 +925,7 @@ public class CognitoUser {
      * @return Valid, cached tokens {@link CognitoUserSession}. {@code null}
      *         otherwise.
      */
-    protected CognitoUserSession getCachedSession() {
+    public CognitoUserSession getCachedSession() {
         synchronized (GET_CACHED_SESSION_LOCK) {
             if (userId == null) {
                 throw new CognitoNotAuthorizedException("User-ID is null");


evstep:
hi there! why this small change delayed for so long?
@kvasukib @desokroshan

androidsudo:
hi there! why this small change delayed for so long?
@desokroshan @kvasukib

kvasukib:
@androidsudo What's your use-case for using this method from your app?

androidsudo:
@kvasukib i developing few app which will use one shared account. For it i'm developing special app which is in charge to handle all authentication stuff as advised by google [here](https://developer.android.com/training/id-auth/custom_auth#DistributeService).
in this app i'm implement [AbstractAccountAuthenticator](https://developer.android.com/reference/android/accounts/AbstractAccountAuthenticator.html) which is have getAuthToken method, that's why it's convinient to have such method `public`.
 Otherwise will have there either some countdown latches for `public void getSession(final AuthenticationHandler callback) `  but i'm will prefer just can call `public CognitoUserSession getCachedSession()`
I hope you understand my point!

androidsudo:
@kvasukib  is there any chance that this PR will be merged?

bryannevega:
Still open in 2020?

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
```diff
@@ -925,7 +925,7 @@ public Runnable respondToMfaChallenge(final String mfaCode,
      * @return Valid, cached tokens {@link CognitoUserSession}. {@code null}
      *         otherwise.
      */
-    protected CognitoUserSession getCachedSession() {
+    public CognitoUserSession getCachedSession() {
```
jamesonwilliams, line 928: @androidsudo @bryannevega - How are you planning to consume this?
```
public boolean isSignedIn() {
    try {
        getCachedSession();
        return true;
    } catch (AmazonClientException sessionAcessFailure) {
        return false;
    }
}
```
Something like that? Would it be more useful to you if this method were cooked into the `CognitoUser`, so you didn't have to introduce it to your consuming code base(s)?

jamesonwilliams:
@bryannevega @androidsudo 

> Still open in 2020?

Sorry for the outrageous response time on this one. Definitely a :pensive: or even a :disappointed: emoji kinda situation, here.

jamesonwilliams:
PR is almost one year old; closing as stale.

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
androidsudo: @jamesonwilliams hello! sorry that missed your question! From the October 2019 i'm using own fork of cognito sdk,
How i use it - mostly as described [here](https://github.com/aws-amplify/aws-sdk-android/pull/1087#issuecomment-542445186) because of async nature of `public void getSession(final AuthenticationHandler callback)` some use cases (for example when i need to have JWT in some background tasks immediately) are requires to have syncing mechanics that's leads to code complexity increasing

PR #1089, opened 2019-07-25
Fixed response unmarshalling with gzip encoding

palpatim:
Fixed response unmarshalling when response is gzip encoded without a CRC32
checksum. Also fixes bug decoding Kinesis responses with GZIP encoding.

Additionally:
- Removed internal reference
- Removed internal reference
- Added missing modules to modules.xml

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/http/JsonResponseHandler.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/http/JsonResponseHandler.java
@@ -86,7 +86,6 @@ public class JsonResponseHandler<T> implements HttpResponseHandler<AmazonWebServ
         log.trace("Parsing service response JSON");
 
         final String crc32Checksum = response.getHeaders().get("x-amz-crc32");
-        CRC32ChecksumCalculatingInputStream crc32ChecksumInputStream = null;
 
         // Get the raw content input stream to calculate the crc32 checksum on
         // gzipped data.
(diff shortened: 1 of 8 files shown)


aws-android-sdk-core/src/main/java/com/amazonaws/http/JsonResponseHandler.java
```diff
@@ -129,7 +136,7 @@ public JsonResponseHandler(Unmarshaller<T, JsonUnmarshallerContext> responseUnma
 
             awsResponse.setResult(result);
 
-            final Map<String, String> metadata = new HashMap<String, String>();
+            final Map<String, String> metadata = new HashMap<>();
```
kvasukib: This change has to be reverted as our maven compile and install step uses Java 1.6 source compatibility and this would fail even though Android Studio Gradle Build succeeds. 

I will look into documenting this and probably adding a build step to our build.gradle to prevent accepting this suggestion from Android Studio.

aws-android-sdk-ddb-mapper-test/src/androidTest/java/com/amazonaws/mobileconnectors/dynamodbv2/dynamodbmapper/BatchLoadIntegrationTest.java
```diff
@@ -217,6 +218,41 @@ public void testExponentialBackOffForBatchGetInMapper()
...
+        Map<String, List<Object>> response = null;
+        itemsToGet.put(clazz, keyPairs);
+        response = gzippingMapper.batchLoad(itemsToGet);
+        List<Object> items = response.get(TABLE_NAME);
+        assertEquals(numItems, items.size());
+
+        for (Object item : items) {
+            objs.contains(item);
+        }
+        Thread.sleep(1000 * 10);
```
kvasukib: Is this sleep necessary?

palpatim: I'm not actually sure--I cargo-cult copy/pasted it from the previous test :)

I'm inclined to leave it in for now, since it's reliable in the pipelines (which is why I assume it was added to begin with).

aws-android-sdk-core/src/main/java/com/amazonaws/http/JsonResponseHandler.java
palpatim: Ah, thanks! Will fix.

kvasukib:
@palpatim Can you add a `CHANGELOG` statement?

Issue #1103, opened 2019-08-01
Upload Fails with error Log tag "AwsChunkedEncodingInputStream" exceeds limit of 23 characters


lionheart1060:
**State your question**
Getting error while uploading any file from Android 6.0.1

**Which AWS Services are you utilizing?**  - S3

**Provide code snippets (if applicable)**
08-01 09:39:43.977 8286-9177/? D/UploadTask: Failed to upload: 2 due to Log tag "AwsChunkedEncodingInputStream" exceeds limit of 23 characters
<img width="890" alt="螢幕快照 2019-08-01 上午10 28 02" src="https://user-images.githubusercontent.com/53545072/62258900-59931580-b447-11e9-8b48-4828967199b0.png">

**Environment(please complete the following information):**
 - SDK Version: 2.13.1

**Device Information (please complete the following information):**
 - Device : LENOVO Lenovo PB2-690M
 - Android Version: [e.g. Marshmallow 6.0.1]
 - Specific to simulators: No

desokroshan:
@lionheart1060 Thanks for reporting the problem. Can you please work off of the [PR](https://github.com/aws-amplify/aws-sdk-android/pull/1105) and let us know if it fixes your issue?

lionheart1060:
@desokroshan Thanks for reply
I'm kinda new to android studio, i should be using sdk version : 2.14.2 on branch roshan/Issue#1103 
? But i can't find the newest version though...<img width="937" alt="螢幕快照 2019-08-02 下午2 37 24" src="https://user-images.githubusercontent.com/53545072/62347174-caaff700-b533-11e9-8873-5595ee47a19b.png">

desokroshan:
Version 2.14.2 isn't published yet so you will not be able to find it in Android Studio. You can do the following to use the PR code in you app : 

1. In a freshly cloned repo, checkout the branch "roshan/Issue#1103"
2. Import the "aws-android-sdk-s3" module by File -> New -> Import Module..., provide path to `aws-android-sdk-s3` module from the branch above and follow the dialog instructions to add the module. You should now be able to see module called `aws-android-sdk-s3` in the project navigator on the left-hand side.
3.  Add dependency on the newly added module as `api project(':aws-android-sdk-s3')` in the app's build.gradle

You should now be able to use the source code in PR in your app. Let me know if you have issues.

lionheart1060:
Thanks a lot, it worked!

desokroshan:
[...]

Issue #787, opened 2019-03-14
Upload Fails with error Log tag "RepeatableFileInputStream" exceeds limit of 23 characters


samyakjain:
**Describe the bug**
Getting error while uploading any file from Android 7.1.1

**Which AWS service(s) are affected?** - S3

**Expected behavior**

**Logs**
019-03-14 13:44:37.645 18186-18517/? E/UploadTask: Error encountered during multi-part upload: 9 due to Log tag "RepeatableFileInputStream" exceeds limit of 23 characters
    
    java.lang.IllegalArgumentException: Log tag "RepeatableFileInputStream" exceeds limit of 23 characters
    
        at android.util.Log.isLoggable(Native Method)
        at com.amazonaws.logging.AndroidLog.a(AndroidLog.java:40)
        at com.amazonaws.services.s3.internal.RepeatableFileInputStream.mark(RepeatableFileInputStream.java:105)
        at com.amazonaws.internal.SdkFilterInputStream.mark(SdkFilterInputStream.java:97)
        at com.amazonaws.util.LengthCheckInputStream.mark(LengthCheckInputStream.java:107)
        at com.amazonaws.internal.SdkFilterInputStream.mark(SdkFilterInputStream.java:97)
        at com.amazonaws.http.AmazonHttpClient.b(AmazonHttpClient.java:306)
        at com.amazonaws.http.AmazonHttpClient.a(AmazonHttpClient.java:212)
        at com.amazonaws.services.s3.AmazonS3Client.a(AmazonS3Client.java:4659)
        at com.amazonaws.services.s3.AmazonS3Client.a(AmazonS3Client.java:1752)
        at com.amazonaws.mobileconnectors.s3.transferutility.UploadTask.b(UploadTask.java:221)
        at com.amazonaws.mobileconnectors.s3.transferutility.UploadTask.call(UploadTask.java:15099)
        at java.util.concurrent.FutureTask.run(FutureTask.java:237)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1133)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:607)
        at java.lang.Thread.run(Thread.java:761)

**Environment Information (please complete the following information):**
 - AWS Android SDK Version: 2.8.4
 - Android Version: [e.g. Nougat 7.1.1]
 - Specific to simulators: No

**Additional context**

[...]

PR #1105, opened 2019-08-01
Shorten log tag length

desokroshan:
*Issue #, if available:*
Fixes https://github.com/aws-amplify/aws-sdk-android/issues/1103
*Description of changes:*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,10 @@
 - Model updates for the following services
   - Amazon Polly
 
+### Bug Fixes
+- **Amazon S3**
+  - Shorten `AwsChunkedEncodingInputStream` log tag to be within the 23 character limit imposed by android on certain API Levels. See [issue #1103](https://github.com/aws-amplify/aws-sdk-android/issues/1103)
+
 ## [Release 2.14.1](https://github.com/aws/aws-sdk-android/releases/tag/release_v2.14.1)
 
 ### New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Release 2.15.0](https://github.com/aws/aws-sdk-android/releases/tag/release_v2.15.0)
 
+### Bug Fixes
+
+- **AWS Core Runtime**
+  - Update `LogFactory.getLog` tot automatically truncate the log tag to be within 23 character limit imposed by Android on certain API Levels. See [issue #1103](https://github.com/aws-amplify/aws-sdk-android/issues/1103)
+
 ### Mis. Updates
 
 - **Breaking Changes**
@@ -23,10 +28,6 @@
 
 - Model updates for the following services
   - Amazon Polly
-
-### Bug Fixes
-- **Amazon S3**
-  - Shorten `AwsChunkedEncodingInputStream` log tag to be within the 23 character limit imposed by android on certain API Levels. See [issue #1103](https://github.com/aws-amplify/aws-sdk-android/issues/1103)
 
 ## [Release 2.14.1](https://github.com/aws/aws-sdk-android/releases/tag/release_v2.14.1)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Bug Fixes
 
 - **AWS Core Runtime**
-  - Update `LogFactory.getLog` tot automatically truncate the log tag to be within 23 character limit imposed by Android on certain API Levels. See [issue #1103](https://github.com/aws-amplify/aws-sdk-android/issues/1103)
+  - Update `LogFactory.getLog` to automatically truncate the log tag to be within 23 character limit imposed by Android for Nougat(7.0) releases and prior(API <= 23). See [issue #1103](https://github.com/aws-amplify/aws-sdk-android/issues/1103)
 
 ### Mis. Updates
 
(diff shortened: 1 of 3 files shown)


palpatim:
Is this the right solution for this? We have many tags that exceed 23 characters, and it seems odd that it's only happening on an upload--why haven't we heard this before from other classes with long tag names?

desokroshan:
> Is this the right solution for this? We have many tags that exceed 23 characters, and it seems odd that it's only happening on an upload--why haven't we heard this before from other classes with long tag names?

This restriction is only on certain Android API levels which is why we might not have heard a lot of instances of this error. But we did have a very similar issue before as well - https://github.com/aws-amplify/aws-sdk-android/issues/787. 
As for other classes I was considering looking for all such classes and truncating the log tag length where it does exceed the said length limit.

palpatim:
If it's an API-level restriction, we'll want a mechanism to ensure we don't hit this condition, rather than a manual scan and hoping that we remember when we create new tag names.

aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
```diff
@@ -35,29 +35,42 @@
      * @return logger
      */
     public static synchronized Log getLog(Class clazz) {
-        return getLog(clazz.getSimpleName());
+
+        String logTag = clazz.getSimpleName();
+        /** Truncate if length is greater than 23. This is to satisfy the restriction
+        imposed by android on API level 23 through 26 of maximum log tag length of 23. */
+        if (logTag.length() > 23) {
+            android.util.Log.w(TAG, "Truncating log tag length as it exceed 23, the limit imposed by android on certain API Levels");
```
palpatim: nit: s/android/Android/

aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
```diff
@@ -35,29 +35,42 @@
...
      * Get the logger for the string tag
-     * @param string the string tag
+     * @param logTag the string tag
      *
      * @return logger
      */
-    public static synchronized Log getLog(final String string) {
-        Log log = logMap.get(string);
+    public static synchronized Log getLog(String logTag) {
+        if (logTag.length() > 23) {
```
palpatim: Consider replacing both of these with a private utility function to `getSafeLogTag` that will truncate (and warn) and could potentially be extended in future to do other sanitizations. That way your code reads more clearly:

```java
String safeTag = getSafeLogTag(logTag); // Or `sanitizeTag` or `normalizeTag`, etc
...
```

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## [Release 2.15.0](https://github.com/aws/aws-sdk-android/releases/tag/release_v2.15.0)
 
+### Bug Fixes
+
+- **AWS Core Runtime**
+  - Update `LogFactory.getLog` tot automatically truncate the log tag to be within 23 character limit imposed by Android on certain API Levels. See [issue #1103](https://github.com/aws-amplify/aws-sdk-android/issues/1103)
```
kvasukib: What is tot?
Can you mention the API levels?

aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
desokroshan: Done

aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
desokroshan: Done

aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
```diff
@@ -74,4 +76,24 @@ private static boolean checkApacheCommonsLoggingExists() {
...
+     */
+    private static String getTruncatedLogTag(String logTag) {
+        if (logTag.length() > 23) {
+            if (checkApacheCommonsLoggingExists()) {
+                Log log = new ApacheCommonsLogging(TAG);
+                log.warn("Truncating log tag length as it exceed 23, the limit imposed by Android on certain API Levels");
+            } else {
+                android.util.Log.w(TAG, "Truncating log tag length as it exceed 23, the limit imposed by Android on certain API Levels");
+            }
+            logTag  =logTag.substring(0, 23);
```
palpatim: nit: spacing around `=`

aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
```diff
@@ -74,4 +76,24 @@ private static boolean checkApacheCommonsLoggingExists() {
...
+    /**
+     * Truncate log tag to be within 23 characters in length as required by Android on certain API levels.
+     *
+     * @param logTag Log tag to be truncated
+     * @return truncated log tag
+     */
+    private static String getTruncatedLogTag(String logTag) {
+        if (logTag.length() > 23) {
+            if (checkApacheCommonsLoggingExists()) {
+                Log log = new ApacheCommonsLogging(TAG);
```
palpatim: Would this break for untruncated tags (like if our LogFactory class was subclassed or renamed to be greater than 23?

desokroshan: It should not until someone renames this class or manually modifies the `TAG` property in this class. It should not break for subclasses of LogFactory with longer log tags since the `TAG` property of LogFactory is immutable.

PR #1116, opened 2019-08-05
Fix GetObject tests to use SigV4 constructor

palpatim:
*Issue #, if available:*

For backwards compatibility, the  no-arg constructor for `AmazonS3Client` uses SigV2 signing unless overridden. As this signing method is deprecated and on a removal path by the S3 service, we want to migrate our integration tests to not use it. 

We made this migration initially, but forgot a couple of instances of the no-arg constructor that were in use by tests. This change migrates documentation and one of the missed tests.

*Description of changes:*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Change Log - AWS SDK for Android
 
+## [Release 2.15.0](https://github.com/aws/aws-sdk-android/releases/tag/release_v2.15.0)
+
+### Mis. Updates
+
+- **Breaking Changes**
+  - Removed deprecated SDKGlobalConfiguration options:
+    - `ENABLE_S3_SIGV4_SYSTEM_PROPERTY`
+    - `ENFORCE_S3_SIGV4_SYSTEM_PROPERTY`
+
 ## [Release 2.14.2](https://github.com/aws/aws-sdk-android/releases/tag/release_v2.14.2)
 
 ### New Features
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ private class S3Example extends AsyncTask<Void,Void,Void>{
                 Regions.SELECT_YOUR_REGION // Region enum
             );
 
-            AmazonS3Client s3Client = new AmazonS3Client(credentialsProvider);
+            AmazonS3Client s3Client = new AmazonS3Client(credentialsProvider, Regions.getRegion(Regions.SELECT_YOUR_REGION));
             File fileToUpload = YOUR_FILE;
             //(Replace "MY-BUCKET" with your S3 bucket name, and "MY-OBJECT-KEY" with whatever you would like to name the file in S3)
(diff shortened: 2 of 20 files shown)


desokroshan:
Changes look good but there seems to be other usages of the `AmazonS3Client` default constructor and `AmazonS3Client(credentialsProvider)` constructor such as the following which do not have region in them : 

https://github.com/aws-amplify/aws-sdk-android/blob/7ba0f6f9e14e87708cd68d0821d7991c51016124/aws-android-sdk-s3/src/test/java/com/amazonaws/services/s3/Amazons3ClientTest.java#L50

https://github.com/aws-amplify/aws-sdk-android/blob/9a8021bea46a8503f4e67c782146924fd1f75a56/aws-android-sdk-s3/src/test/java/com/amazonaws/services/s3/TransferUtilityTest.java#L67

https://github.com/aws-amplify/aws-sdk-android/blob/7ba0f6f9e14e87708cd68d0821d7991c51016124/aws-android-sdk-s3/src/test/java/com/amazonaws/services/s3/Amazons3ClientTest.java#L196

Are these overridden to use SigV4 and not need migration?
Note : Above is not an exhaustive list of the occurrences of said constructors.

kvasukib:
@desokroshan The ones you pointed out are unit tests, so they don't make a call to S3. However, we can change it to use the constructor with the region.

We should be checking for *IntegrationTest.java, if there is one that uses the constructor without a region.

desokroshan:
Following are some of the integration tests that use the prohibited constructor : 

https://github.com/aws-amplify/aws-sdk-android/blob/881bb946777a56488c659c8ef8c210d1bf689f3d/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/S3SkipByteDigestIntegrationTest.java#L48

https://github.com/aws-amplify/aws-sdk-android/blob/881bb946777a56488c659c8ef8c210d1bf689f3d/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/S3InterruptIntegrationTest.java#L54

https://github.com/aws-amplify/aws-sdk-android/blob/881bb946777a56488c659c8ef8c210d1bf689f3d/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/model/GeneratePresignedUrlRequestIntegrationTest.java#L63

Also, there are integration tests other S3 which use Amazon S3 client constructor without region such as the following : 

https://github.com/aws-amplify/aws-sdk-android/blob/3bf480cbe82117d67101783267093b8049f90ed5/aws-android-sdk-ddb-mapper-test/src/androidTest/java/com/amazonaws/mobileconnectors/dynamodbv2/dynamodbmapper/S3ClientCacheIntegrationTest.java#L60

palpatim:
Thanks for the callout--I should have checked across the entire test base for uses of the deprecated constructor.

Issue #889, opened 2019-04-09
Unable to verify secret hash for client in v2.13.1 too but working fine for iOS


usmanrana07:
I'm implementing **CustomAuthChallenge**'. This issue occurs when:

-  I try to pass user name for getting user session which is lets say 'abc_xyz'
-  Then it comes in '`getAuthenticationDetails`' method of the callback where i set username and auth type.
-  Then '`authenticateChallenge`' method is called in which 'continuation' object is received which has  a username in its params and that username is not the same that i passed but is userId '12121' of that user. 
- I set ANSWER in 'authenticateChallenge' and let it continueTask
- '`onFailure`' is called with the error 'Unable to verify secret hash for client'

 I checked the sdk and i saw  that userId is treated as '**usernameInternal**' . If i use userId '12121' in replacement of username 'abc_xyz' instead then it works fine. But in iOS it is working with username and we don't want to disclose userId so using username is mandatory. How can i resolve this issue?

usmanrana07:
when is this expected to be resolved?

charandas:
@usmanrana07 what's interesting is that `secretHash` can be calculated with `userId` for `getAuthenticationDetails`, it allows you to override (not that you need this capability):

```
           override fun getAuthenticationDetails(authContinuation: AuthenticationContinuation, username: String) {
                val authParameters = mutableMapOf<String, String>()
                authParameters[CognitoServiceConstants.AUTH_PARAM_USERNAME] = username
                val hash = CognitoSecretHash.getSecretHash(
                    authParams?.userId,
                    CognitoUserPoolHelper.CLIENT_ID,
                    CognitoUserPoolHelper.CLIENT_SECRET
                )

                val authenticationDetails = AuthenticationDetails(authParams?.userId, authParameters, null)
                authenticationDetails.authenticationType = CognitoServiceConstants.AUTH_TYPE_INIT_CUSTOM_AUTH
                authenticationDetails.customChallenge = CognitoServiceConstants.CHLG_TYPE_CUSTOM_CHALLENGE
                authenticationDetails.setAuthenticationParameter(CognitoServiceConstants.AUTH_PARAM_SECRET_HASH, hash)
                authenticationDetails.setAuthenticationParameter(CognitoServiceConstants.CHLG_RESP_SECRET_HASH, hash)
                authContinuation.setAuthenticationDetails(authenticationDetails)
                authContinuation.continueTask()
            }
```

[...]

PR #1144, opened 2019-08-21
[finishes #167549816] User internal username for computing secret hash that is passed in challenge response

desokroshan:
*Issue #, if available:*
Fixes https://github.com/aws-amplify/aws-sdk-android/issues/889

*Description of changes:*
Challenge Response that is passed to cognito with a call to [`respondToAuthChalllenge`](https://github.com/aws-amplify/aws-sdk-android/blob/ada672f3d2dc432d7979c33e5afd4fdfbb2be9ba/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/services/cognitoidentityprovider/AmazonCognitoIdentityProviderClient.java#L4835) needs username and a secretHash if client secret was generated for the App client. There are two different usernames that are being used in the SDK. There is an external username or one that is supplied by the user. There is an internal username which corresponds to the external username. SecretHash must be computed using the username that is passed to the challenge response. In other words, username and secret hash must correspond. For the custom auth flow, username being passed was internal username(aka UserSub) whereas the secretHash was computed using the external username provided by the user(the user facing username). This led to failure in hash verification. This PR addresses the inconsistency.


desokroshan:
> LGTM. Please verify that your test fails without your change. Assuming it does, ship it

I have verified that it does. Thanks!

Issue #888, opened 2019-04-09
Authentication by Cognito Hosted UI can not get token if my app killed


HayTran94:
**Describe the bug**
I am using Hosted UI `AWSMobileClient.getInstance().showSignIn` API to authenticate my users. I can not get user token if my app killed. 

**To Reproduce**
Login by `AWSMobileClient.getInstance().showSignIn` API, the app can get token, identityId and userState is `SIGNED_IN`.
Kill the app and open again, the app can not get token (no callback return), but it still can get identityId and userState is `SIGNED_IN`. Even the app call `getTokens` many times, but no callback invoked. However, if I invoke `signOut` API,  the `onError` callback of `getTokens` will be invoked with `java.lang.Exception: getTokens does not support retrieving tokens while signed-out`.

**Which AWS service(s) are affected?**
AWSMobileClient 

**Expected behavior**
The app can get user access token as long as user session is valid (refresh token is valid) even if the app killed. 

**Screenshots**
The following logs show logs after invoke `getTokens` and `getIdentityId`. As you can see, no `getTokens` callback return, but `getIdentityId` return a valid Id. 
<details> 
<summary>Click me to see logs</summary>

```
2019-04-09 12:28:12.603 21157-21174/com.abc.abc D/AWSMobileClient: Inspecting user state details
2019-04-09 12:28:12.605 21157-21174/com.abc.abc I/SplashScreenActivity: Result: SIGNED_IN
2019-04-09 12:28:16.254 21157-21208/com.abc.abc D/AWSMobileClient: Inspecting user state details
2019-04-09 12:28:16.364 21157-21208/com.abc.abc W/AWSMobileClient: Tokens are invalid, please sign-in again.
    java.lang.NullPointerException: Attempt to invoke virtual method 'void com.amazonaws.mobileconnectors.cognitoauth.Auth.setAuthHandler(com.amazonaws.mobileconnectors.cognitoauth.handlers.AuthHandler)' on a null object reference
        at com.amazonaws.mobile.client.AWSMobileClient$8.run(AWSMobileClient.java:1555)
        at com.amazonaws.mobile.client.internal.InternalCallback.await(InternalCallback.java:115)
        at com.amazonaws.mobile.client.AWSMobileClient.getTokens(AWSMobileClient.java:1517)
        at com.amazonaws.mobile.client.AWSMobileClient.getUserStateDetails(AWSMobileClient.java:1011)
[...]

Issue #873, opened 2019-04-05
Tokens are invalid, please sign-in again after call AWSMobileClient.getInstance().getTokens(callback)


tvanmanh:
**Describe the bug**
I have already signed in and then I kill app and open app again.
In my plash screen I call API   AWSMobileClient.getInstance().initialize()  and check the user state is  "SIGNED_IN".
After that I go to next step, call my backend API which needs token using  AWSMobileClient.getInstance().getTokens(new Callback<Tokens>()) after i call that API. I got user state is " SIGNED_OUT_USER_POOLS_TOKENS_INVALID" and This call blocked the thread forever don't return token or any error. 

I know that the token will expire after 1 hour but I have already signed in. Just kill my app and open app again How can i got SIGNED_OUT_USER_POOLS_TOKENS_INVALID after call get token. 

I resolved it by add UserStateListener after getting SIGNED_OUT_USER_POOLS_TOKENS_INVALID i will sign in again but should I do that every time I kill app and open again ?

**To Reproduce**

<details><summary>Code and logs</summary>

```
`My code in plash screen: 
AWSMobileClient.getInstance().initialize(getApplicationContext(), new Callback<UserStateDetails>() {
                    @Override
                    public void onResult(UserStateDetails userStateDetails) {
                        Log.i("MYLOG", "Result: " + userStateDetails.getUserState());
                        initializeSingletonClass();

                        if (userStateDetails.getUserState() == UserState.SIGNED_IN) {
                            routeToAppropriatePage(ActMain.class);
                        } else {
                            routeToAppropriatePage(ActLoginOption.class);
                        }
                    }

                    @Override
                    public void onError(Exception e) {
                        Log.e(TAG, "MYLOG", e);
                        routeToAppropriatePage(ActLoginOption.class);
                    }
                }
        );
        AWSMobileClient.getInstance().addUserStateListener(new UserStateListener() {
            @Override
            public void onUserStateChanged(UserStateDetails details) {
                switch (details.getUserState()) {
                    case GUEST:
                        Log.i("MYLOG", "user is in guest mode");
                        break;
                    case SIGNED_OUT:
                        Log.i("MYLOG", "user is signed out");
                        break;
                    case SIGNED_IN:
                        Log.i("MYLOG", "user is signed in");
                        break;
                    case SIGNED_OUT_USER_POOLS_TOKENS_INVALID:
                        Log.i("MYLOG", "need to login again");
                        break;
[...]

PR #1150, opened 2019-08-22
[AWSMobileClient] Fix the object lifecycle of hostedUI

kvasukib:
*Issue #, if available:*

#888
#873

*Description of changes:*

The changes include expanding the lifecycle of `hostedUI` object to be created when `initialize` is called and not destroy it when `signOut` is called.


palpatim:
For future reference: We recognize that this fix is a bit of a hack; we're setting a default value for hostedUI upon Mobile Client init and grabbing the latest username so that a common app flow of Sign In -> Quit app -> Launch App -> Initialize MobileClient -> getTokens doesn't cause a crash.

Longer term, we want to abstract the retrieval of tokens from the UI used to present sign in to the user, but that is complex enough that we want to get this rather urgent fix out first.

PR #1170, opened 2019-08-30
Added exception handler for getCurrentUser in CognitoUserPool

alex-Lotus:
*Description of changes:* Added exception handler for CognitoUserPool when getting current user and  javax.crypto.AEADBadTagException or other crash occurs

### DIFF
--- a/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUserPool.java
+++ b/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUserPool.java
@@ -484,11 +484,15 @@ public class CognitoUserPool {
         final String csiLastUserKey = "CognitoIdentityProvider." + clientId + ".LastAuthUser";
 
         if (awsKeyValueStore.contains(csiLastUserKey)) {
-            return getUser(awsKeyValueStore.get(csiLastUserKey));
-        } else {
-            return getUser();
+            try {
+                return getUser(awsKeyValueStore.get(csiLastUserKey));
+            } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUserPool.java
```diff
@@ -484,12 +484,16 @@ public CognitoUser getCurrentUser() {
...
 
         if (awsKeyValueStore.contains(csiLastUserKey)) {
-            return getUser(awsKeyValueStore.get(csiLastUserKey));
-        } else {
-            return getUser();
+            try {
+                return getUser(awsKeyValueStore.get(csiLastUserKey));
+            } catch (Exception e) {
+                //  Exception like javax.crypto.AEADBadTagException
+            }
```
jamesonwilliams, line 491: Couple of thoughts on this change -
1. Generally speaking, `catch`ing the base `Exception` type [is not a best practice](https://stackoverflow.com/a/2416334/695787); should prefer more specific exception type, here;
2. Empty `catch` blocks are [not a best practice](https://stackoverflow.com/a/1234364/695787)

While reading this code I do notice also that the `contains`/`get` calls on the key-store are race-y. Might be better as:
```
final String csiLastUserValue;
try {
    csiLastUserValue = awsKeyValueStore.get(csiLastUserKey);
} catch (GeneralSecurityException keyStoreAccessFailure) {
    // Either wrap the GSE into an existing Cognito error type, or, possibly, return null:
    return null;
}
if (csiLastUserValue != null) {
    return getUser(csiLastUserValue);
}
return getUser();
```

jamesonwilliams:
Closing PR as stale. Please re-open if you get time to address the feedback! Thanks

PR #1176, opened 2019-09-04
Verify also if clientSecret isn't an empty string avoid UnsupportedOperationException

adrianolc:
I'm trying to create CognitoUserPool using the awsconfiguration.json file passing the instance of AWSConfiguration instead of creating CognitoUserPool passing all ids on the constructor.

With this constructor 
`/**
     * Constructs a user-pool with default {@link ClientConfiguration}.
     *
     * @param context               REQUIRED: Android application context.
     * @param awsConfiguration      REQUIRED: Holds the configuration read from awsconfiguration.json
     */
    public CognitoUserPool(Context context, AWSConfiguration awsConfiguration) `

Our pool doesn't have the AppClientSecret so we don't pass it on the awsconfiguration file, but this constructor will try to read the property and add an empty string case it not found the property.

`this.clientSecret = userPoolConfiguration.optString("AppClientSecret");`

Then when we go to the initiateCustomAuthRequest it throws an exception when try to add a null or empty secret hash

*Issue #, if available:*

*Description of changes:*

Verify also if the string is not empty

### DIFF
--- a/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
+++ b/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
@@ -2921,7 +2921,7 @@ public class CognitoUser {
          * the secret hash is not passed in.
          */
         Map<String, String> authenticationParameters = authenticationDetails.getAuthenticationParameters();
-        if (clientSecret != null &&
+        if (clientSecret != null && !clientSecret.isEmpty() &&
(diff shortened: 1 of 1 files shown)


desokroshan:
@adrianolc Thanks for the PR. Interestingly, I don't run into any issue trying to use a configuration file without client secret. I used the same constructor to initialize the user pool as follows :

```cup = new CognitoUserPool(getApplicationContext(), new AWSConfiguration(getApplicationContext()));```.  Do you mind opening a linked issue with a detailed stack trace and relevant code snippet for further investigation?

desokroshan:
Closing per discussion in https://github.com/aws-amplify/aws-sdk-android/issues/1191

PR #1181, opened 2019-09-04
Update README.md to remove suggestion on Dynamic versioning of dependencies

desokroshan:
*Issue #, if available:*

*Description of changes:*
Dynamic versioning of dependencies is not a recommended practice and causes a warning from Android Studio. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,7 @@ Simply add to your app's build.gradle (Module)'s dependencies{ } section
 
 SERVICE can be `s3`, `ddb`, `pinpoint`, etc.
 
-(hint: you can specify 2.+ or 2.x.+ to automatically get the latest version where the + is specified.
-Remember for version 2.X.Y an increase in X may indicate a breaking change or new features, while increases in Y are likely bug fixes, or small feature improvements)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,8 +49,7 @@ Simply add to your app's build.gradle (Module)'s dependencies{ } section
 
 SERVICE can be `s3`, `ddb`, `pinpoint`, etc.
 
-(hint: you can specify 2.+ or 2.x.+ to automatically get the latest version where the + is specified.
-Remember for version 2.X.Y an increase in X may indicate a breaking change or new features, while increases in Y are likely bug fixes, or small feature improvements)
+(Note: Remember for version 2.X.Y an increase in X may indicate a breaking change or new features, while increases in Y are likely bug fixes, or small feature improvements)
```
palpatim: Do we need this at all, since we have https://aws-amplify.github.io/docs/android/manualsetup#aws-sdk-version-vs-semantic-versioning ?

jamesonwilliams: During Bug Bash, we found this old PR and we decided to include it. We will come back and make a note about using Semantic Versioning, also. This current PR includes a useful update, removing our suggestion to use the `+` notation on versions.

Issue #1006, opened 2019-06-04
Video upload fails to Amazon S3 bucket inconsistently


partheebanj:
**Describe the bug**
Uploading videos to Amazon S3 fails inconsistently. I have been able to upload videos successfully, but there are certain instances when the upload fails with the following error: com.amazonaws.AmazonClientException: Unable to execute HTTP request: connect timed out. 

**To Reproduce**
Uploading videos to Amazon S3

Your code

                val key = AccountSessionManager.awsPath(remoteKey, bucket,t)
                val bucketName = AccountSessionManager.awsBucket(bucket,t)

                val clientConfiguration = ClientConfiguration()
                        .withMaxErrorRetry(3) // 3 retries
                        .withConnectionTimeout(120000) // 120,000 ms
                        .withSocketTimeout(120000) // 120,000 ms

                val credentials = BasicSessionCredentials(t!!.accessKeyId, t.secretAccessKey, t.sessionToken)
                val s3Client = AmazonS3Client(credentials, Constants.S3Region, clientConfiguration)

                val transferUtility = TransferUtility
                        .builder()
                        .s3Client(s3Client)
                        .context(mContext)
                        .build()

                val uploadObserver = transferUtility.upload(bucketName, key, file)

**Which AWS service(s) are affected?**
Amazon S3

**Expected behavior**
Video upload successfully without any failures

 - AWS Android SDK Version: 2.13.0
 - Device: Samsung
 - Specific to simulators: No

The internet connection speed is good and I have been able to successfully upload many videos except some. Trying to upload again works fine, but want to know the root cause on why I am getting this error.

kvasukib:
@partheebanj Thank you for reporting to us. The exception you reported occurs when the network goes offline when a transfer is in progress. We have two recommendations to checkpoint the transfer and then resume when network comes back online.

1) Using a `TransferService` will register a callback with Android to get notified of the network connectivity changes. When network goes offline, the ongoing transfers are paused and when the network comes back online, the ongoing transfers are resumed. `TransferService` can handle this irrespective of the app being in background or foreground.

2) Using a `TransferNetworkLossHandler`: You can create an object of this handler and register it in your application. By doing so, when network goes offline, the ongoing transfers are paused and when the network comes back online, the ongoing transfers are resumed. `TransferNetworkLossHandler` only works when the app is in foreground. To make this work when your app is in background, you need to wrap the creation of this handler in an Android component (like a Service, etc.)

[...]

PR #1185, opened 2019-09-06
Issue#1006

raphkim:
*Issue #, if available:*
[1006](https://github.com/aws-amplify/aws-sdk-android/issues/1006)

*Description of changes:*
The following race condition would cause transfer to fail:
1. Network is lost
1. Transfer is interrupted
1. Network is recovered
1. Transfer task checks network condition
1. Sees that network is connected and updates `TransferState` to FAILED instead of WAITING_FOR_NETWORK

A flag was added to `TransferNetworkLossHandler` singleton to remember that network was lost until transfers are successfully resumed.


raphkim:
did not fix the issue

Issue #1236, opened 2019-10-04
S3 upload MD5 not automatically populated


emarc-m:
**State your question**

I've read the documentation for the MD5 file integrity check when uploading files to S3. Please see: https://aws-amplify.github.io/aws-sdk-android/docs/reference/com/amazonaws/services/s3/model/ObjectMetadata.html#getContentMD5--

It says on the documentation that 
```
The AWS S3 Java client will attempt to calculate this field automatically when uploading files to Amazon S3.
```

However, looking at the database entry for the transfer utility, it does not contain any data about the MD5. Also, checking `UploadTask#createPutObjectRequest L:457`, `upload.md5` is null. 

Does this MD5 header need to be computed and set manually? Thank you.

**Which AWS Services are you utilizing?**

AWS S3

**Provide code snippets (if applicable)**
Java docs snippet:
```java
/**
     * <p>
     * Sets the base64 encoded 128-bit MD5 digest of the associated object
     * (content - not including headers) according to RFC 1864. This data is
     * used as a message integrity check to verify that the data received by
     * Amazon S3 is the same data that the caller sent. If set to null,then the
     * MD5 digest is removed from the metadata.
     * </p>
     * <p>
     * This field represents the base64 encoded 128-bit MD5 digest digest of an
     * object's content as calculated on the caller's side. The ETag metadata
     * field represents the hex encoded 128-bit MD5 digest as computed by Amazon
     * S3.
     * </p>
     * <p>
     * The AWS S3 Android client will attempt to calculate this field
     * automatically when uploading files to Amazon S3.
     * </p>
     *
     * @param md5Base64 The base64 encoded MD5 hash of the content for the
     *            object associated with this metadata.
     * @see ObjectMetadata#getContentMD5()
     */
    public void setContentMD5(String md5Base64) {
        if (md5Base64 == null) {
            metadata.remove(Headers.CONTENT_MD5);
        } else {
            metadata.put(Headers.CONTENT_MD5, md5Base64);
        }

    }
```

**Environment(please complete the following information):**
 - SDK Version: 2.16.+ (Gradle)

**Device Information (please complete the following information):**
 - Device: Android Emulator
 - Android Version: Android 8.1 Google Play

raphkim:
Hi @emarc-m 

To answer your question, the S3 client will automatically set and attach content MD5 hash to object metadata without having to do so manually. Would you be able to provide us with the code snippet you are using to set up your transfer utility object and to upload the file?

[...]

PR #1240, opened 2019-10-08
[S3] content md5 in PutObjectResult

raphkim:
*Issue #, if available:*
#1236 

*Description of changes:*
Added a line to set content MD5 in `PutObjectResult` to match the behavior in [aws-sdk-java](https://github.com/aws/aws-sdk-java/blob/5bafd0d6c4a979bc55fcf6709e93c9983bd2d1e6/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java#L5812).


aws-android-sdk-s3/src/main/java/com/amazonaws/services/s3/AmazonS3Client.java
```diff
@@ -2003,6 +2003,7 @@ public PutObjectResult putObject(PutObjectRequest putObjectRequest)
         result.setETag(returnedMetadata.getETag());
         result.setMetadata(returnedMetadata);
         result.setRequesterCharged(returnedMetadata.isRequesterCharged());
+        result.setContentMd5(contentMd5);
```
kvasukib: can you add unit and/or integration test for contentMd5 to be null, non-null and contentMd5 passed through ObjectMetadata by the developer?

PR #1278, opened 2019-10-27
[aws-android-sdk-pinpoint] Update models to latest

awsmobilesdk:
[aws-android-sdk-pinpoint] Update models to latest


kvasukib:
@raphkim can you check again if the pinpoint model generator is turned off?

Closing the PR as the generated code is not correct.

PR #1292, opened 2019-11-04
Update javadoc to warn against using BasicAWSCredentials

raphkim:
*Issue #, if available:*
N/A

*Description of changes:*
Update javadoc with a link to [AWSMobileClient](https://aws-amplify.github.io/aws-sdk-android/docs/reference/com/amazonaws/mobile/client/AWSMobileClient.html) as recommended alternative to using `BasicAWSCredentials`.

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/auth/BasicAWSCredentials.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/auth/BasicAWSCredentials.java
@@ -18,6 +18,10 @@ package com.amazonaws.auth;
 /**
  * Basic implementation of the AWSCredentials interface that allows callers to
  * pass in the AWS access key and secret access in the constructor.
+ *
+ * It is not recommended to use this class. Instead, use short term credentials
+ * from Cognito with the AWSMobileClient.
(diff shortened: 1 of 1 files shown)


aws-android-sdk-core/src/main/java/com/amazonaws/auth/BasicAWSCredentials.java
```diff
@@ -18,6 +18,10 @@
 /**
  * Basic implementation of the AWSCredentials interface that allows callers to
  * pass in the AWS access key and secret access in the constructor.
+ *
+ * It is not recommended to use this class. Instead, use short term credentials
+ * from Cognito with the AWSMobileClient.
```
kvasukib: It is not recommended to use long-term credentials through this class. Instead, use short-term credentials
 vended from Amazon Cognito through {@link com.amazonaws.mobile.client.AWSMobileClient}.

raphkim: link doesn't work because core SDK is not a dependent on auth-core SDK

Issue #1306, opened 2019-11-15
Passwordless Custom Auth flow


ximenaperez:
Hi all,

I'm having an issue when attempting sign in doing a custom auth flow for passwordless sign in.
I followed the docs here: https://aws-amplify.github.io/docs/android/authentication#custom-authentication-in-amplify

I got a awsconfiguration.json exactly like the one there, with those two sections including the custom part.

My issue is that when I attempt sign in, with a dummy password as it says there, the onError method gets called, with error code InvalidParameterException and message: USER_SRP_AUTH is not enabled for the client.

I assume this is because I'm sending a dummy password, but there is no way to NOT send a password right? and the doc says I should send a dummy password. Is there anything else I'm missing? does anyone know why that error specifically could be happening in a passwordless scenario?

These are my gradle imports:
implementation 'com.amazonaws:aws-android-sdk-core:2.16.3'
implementation 'com.amazonaws:aws-android-sdk-appsync:2.8.+'
implementation 'com.amazonaws:aws-android-sdk-cognitoidentityprovider:2.13.2'

I've only tried on a Pixel 2 with Android 9.

Thanks in advance for the help!

desokroshan:
@ximenaperez Thanks for reaching out. Can you share your user pool settings?

ximenaperez:
Hi @desokroshan 

I have moved forward with this issue a little bit, but still not getting it to work properly. I noticed that I was missing a few dependencies, and after updating my dependencies to the latest versions and adding a user pool dependency, the sign in worked.

My issue now is that the sign in returns SIGNINSTATE = DONE, when it should be CUSTOM_CHALLENGE. After adding logs on the lambda functions, particularly the define one, I was able to compare a web client that is working correctly versus my Android SDK client that it's not (both using the same lambda functions).

The main difference, is that when we log the SESSION object, the web client has an empty session when attempting sign in, like this:

a3083668-289d-4456-8528-d9e8de338691	INFO	[]

but the Android one (same user even) shows this:

064b13bf-9199-47fc-9399-527c1fc05de8	INFO	[ { challengeName: 'SRP_A',
    challengeResult: true,
    challengeMetadata: null } ]

That means that by having that challenge name SRP_A, on the lambda function, it goes through a different (if) than the web client and it does not return the CUSTOM CHALLENGE I want.

[...]

PR #1334, opened 2019-11-26
Add support to Custom Challenge without password check.

adrianolc:
The suggestion change here solve the problem with the issue https://github.com/aws-amplify/aws-sdk-android/issues/1306 

If we create an instance from `AuthenticationDetails` using the constructor with password it going to send a request where the challengeName is `SRP_A` the server is waiting for a request where this name is `CUSTOM_CHALLENGE` that is exactly what happens when we create an instance using the constructor bellow: 

```
    /**
     * Constructs a new object for custom authentication.
     *
     * @param userId REQUIRED: User ID, NOTE: This will over ride the current
     *            Used ID.
     * @param authenticationParameters REQUIRED: Authentication details to
     *            launch custom authentication process.
     * @param validationData REQUIRED: Contains authentication parameters 
     *            which are passed to triggered pre-auth lambda. trigger 
     */
    public AuthenticationDetails(String userId, Map<String, String> authenticationParameters,
            Map<String, String> validationData)
```
 


desokroshan:
@adrianolc Thanks for submitting this PR. But I have tried this before and as you have noted in https://github.com/aws-amplify/aws-sdk-android/issues/1306#issuecomment-564567481, the use of `AuthenticationDetails` constructor without password results in `NotAuthorizedException`. The passwordless custom auth currently requires dummy password and the define auth challenge lambda trigger should have SRP_A as the first challenge as demonstrated in the following docs : https://aws-amplify.github.io/docs/android/authentication#lambda-trigger-setup, which I do agree is unfortunately less than ideal.

PR #1368, opened 2019-12-09
[2.16.5] Add changelog entry

kvasukib:


raphkim:
Can we add KMS and Kinesis Video in here too?

kvasukib:
> Can we add KMS and Kinesis Video in here too?

Added!

PR #1370, opened 2019-12-09
Remove activity related flags in notification intent

kvasukib:
* Removed setting of activity related flags in the notification intent being created. The activity related flags are useful when an activity launches an other activity that is already in the Android activity stack. However, the intent that is being created is not used for communicating with an activity and instead used for service and broadcast. Hence, this comes down to the problem of setting activity related flags in a pending intent which is not used for communicating with an activity.

* A customer found crashes while setting these flags in Android < 4.4. However >= 4.4, Android prevented the crash and made it a no-op. 

* The fix proposed here removes the flags and variable renaming and indentation fixes.


.idea/dbnavigator.xml
```diff
@@ -0,0 +1,454 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
TrekSoft: Probably didn't want to include this file right?

kvasukib: Deleted!

Issue #429, opened 2018-03-23
How to include custom authorization headers while connecting?


SanyasiraoM:
### To help us solve your problem better, please answer the following list of questions.

* What service are you using?
AWS IOT

* In what version of SDK are you facing the problem?
2.6.7

* Is the issue limited to Simulators / Actual Devices?
Both

* Can your problem be resolved if you bump to a higher version of SDK?
No

* Is this problem related to specific Android/OS version?
No

* Can you give us steps to reproduce with a minimal, complete, and verifiable example? Please include any specific network conditions that might be required to reproduce the problem.

This is not an issue. We came to know that we can add custom authorization headers while connecting to aws iot broker at https://docs.aws.amazon.com/iot/latest/developerguide/custom-auth.html. So is there any snippet or reference to how to use them in android and ios?

rohandubal:
Hello,

I have added this as a feature request in our backlog and have submitted it to the team for prioritization. I do not have specific timeline that I can share at this point. We will let you know, either on this thread, or through a forum announcement when the feature is released in the SDK.

Thanks,
Rohan

iamtonny:
Hi, do you have any news about custom authentication for android and ios sdk??? When it can be implemented (month, half-year, year) ??  Thanx

desokroshan:
@iamtonny We are currently working on this request. I cannot share a specific timeline but I will update this thread once the feature has been released.

j0bro:
I am currently working on a big project where this would be very helpful; the SDK documentation claims it allows for setting the custom authorization headers, but the implementation is only present in the JS SDK. In [this forum thread](https://forums.aws.amazon.com/thread.jspa?messageID=836976) there is a post by user vita1ii who created a patch on the Paho MQTT client which is embedded in the AWS Android SDK.
Note that I am only assuming this functionality is also lacking from the iOS SDK, I am just following what was said above here.

kvasukib:
@j0bro We are waiting on Paho to release this fix so we could take a dependency on the version that has the fix. However for iOS SDK, we did not have any limitation, so we have this feature working for iOS: https://aws-amplify.github.io/docs/ios/pubsub#custom-authentication

SandyRagaie:
I am working on Android project now and this feature request blocks a main functionality in the application, is there any known work around we can do to overcome this until aws sdk is released?

desokroshan:
[...]

PR #1457, opened 2020-01-28
[IoT] Add support for custom authentication in IoT connection

desokroshan:
*Issue #, if available:*
Fixes https://github.com/aws-amplify/aws-sdk-android/issues/429

*Description of changes:*

Added support for Custom Authorizers. AWS IoT allows you to define custom authorizers that allow you to manage your own authentication and authorization strategy using a custom authentication service and a Lambda function. Custom authorizers allow AWS IoT to authenticate your devices and authorize operations using bearer token authentication and authorization strategies. See [AWS IoT Custom Authentication](https://docs.aws.amazon.com/iot/latest/developerguide/custom-authentication.html) for more details.


aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -779,20 +805,43 @@ public void onMessageArrived(String topic, byte[] data) {
...
+                public void onStatusChanged(AWSIotMqttClientStatus status, Throwable throwable) {
+                    statuses.add(status);
+                }
+            });
+        } else if (authMode.equals(CUSTOM_AUTH)) {
+            // connect using WebSockets and custom authentication token
+            mqttManager.connect(
+                    "token-header",
+                    "allow",
+                    "Gv85h71GT07RWo5uWQpqHlYWePN7YVY7KgiuPVEnt65+tDDWyZOlGLIDxvwC7LiBF7FbmG/coeBELKltqmrP0gjM7sIyXlrx315xG3LupoK4bmsNIzg6xxWa0n60WxhOv1p5qXDAhmYWIFuL0enMtuwpCY1W073NqZ/+iOsrJexssZiiLnvVf3z6HReEEER6TfFevrgAEUva0h9pzbUepbJ6RaOO7zZY6e/T/wSHrKl0jA4MalGReKbhdlg0HyTPfILhjorYkF7ew/8Ml+QMr/WU90YMp9Sv8SF0q7WBzbuqxpAR8vxm00rRlgrFdsIRjGFTDmrp9Fsx+xlx0aADCA==",
```
raphkim: let's not have a hard-coded token here

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -723,20 +725,44 @@ public void onFailure(Throwable exception) {
...
-                statuses.add(status);
-            }
-        });
+
+        if (authMode.equals(CUSTOM_AUTH)) {
+            // connect using WebSockets and custom authentication token
+            mqttManager.connect(
+                    "token-header",
+                    "allow",
+                    "Gv85h71GT07RWo5uWQpqHlYWePN7YVY7KgiuPVEnt65+tDDWyZOlGLIDxvwC7LiBF7FbmG/coeBELKltqmrP0gjM7sIyXlrx315xG3LupoK4bmsNIzg6xxWa0n60WxhOv1p5qXDAhmYWIFuL0enMtuwpCY1W073NqZ/+iOsrJexssZiiLnvVf3z6HReEEER6TfFevrgAEUva0h9pzbUepbJ6RaOO7zZY6e/T/wSHrKl0jA4MalGReKbhdlg0HyTPfILhjorYkF7ew/8Ml+QMr/WU90YMp9Sv8SF0q7WBzbuqxpAR8vxm00rRlgrFdsIRjGFTDmrp9Fsx+xlx0aADCA==",
```
raphkim: Please add this into `testconfiguration.json` or use other secure means

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
desokroshan: Couple of thoughts on this : 
1. I don't think token signature is something that has to be obfuscated. It is similar to the AWS signature that we include in all the service calls. It is generated by signing token with the private key and is meant to be "in-transit" and purpose of signing it is to prevent any wrongdoings from a malicious third party that happens to get its hand on the request. IoT verifies the authenticity of the request by using its public key.

2. As far as token itself is concerned, there is not much one can do without the user-specific IoT endpoint. It is akin to https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-mobile-client/src/androidTest/java/com/amazonaws/mobile/client/AWSMobileClientTest.java#L95

Purpose of putting it like this was to emphasize the above point about its usage. I will redact these for now just to be safe and will revisit this later.

raphkim: As per our earlier discussion, I think a better way to reflect example usage would be to show them how this token is obtained with SSL encryption of token and private key. Token should be obtained exactly how a customer would obtain it and private key should be the one being encrypted.

desokroshan: Token is obtained using customer's own custom authentication mechanism and I have added the process for obtaining token signature in the documentation. Sample app would be more appropriate place to show how to do that programmatically instead of the tests.

raphkim: I agree that sample app is a better place for this. Sounds good to me

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -723,20 +725,44 @@ public void onFailure(Throwable exception) {
...
         mqttManager.setAutoReconnect(false);
-        // connect using WebSockets and IAM credentials
-        mqttManager.connect(credentialsProvider, new AWSIotMqttClientStatusCallback() {
-            @Override
-            public void onStatusChanged(AWSIotMqttClientStatus status, Throwable throwable) {
-                statuses.add(status);
-            }
-        });
+
+        if (authMode.equals(CUSTOM_AUTH)) {
```
jamesonwilliams: It's always better to do these like:
```
FOR_SURE_NON_NULL.equals(possiblyNull)
```
When `possiblyNull` _is_ `== null`, then that statement evaluates as false.

However, the statement `possiblyNull.equals(FOR_SURE_NON_NULL)` throws a `NullPointerException` in that same situation.

^^^ Same comment for every place you have an `enumValue.equals(Enum.CONSTANT)`

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -779,20 +805,43 @@ public void onMessageArrived(String topic, byte[] data) {
 
     @Test
     public void mqttWebSocketReconnect() throws Exception {
+        websocketReconnectionTest(IAM);
+    }
 
+    @Test
+    public void testCustomAuthReconnect() throws Exception {
+        websocketReconnectionTest(CUSTOM_AUTH);
```
jamesonwilliams: `testWebsocketReconnectWithIam`, `testWebsocketReconnectWithCustomAuth`

Something like that

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -941,6 +947,72 @@ public void run() {
...
+        LOGGER.debug("MQTT broker endpoint: " + endpointWithHttpPort);
+        try {
+            if (mqttClient == null) {
+                mqttClient = new MqttAsyncClient(
+                        "wss://" + endpointWithHttpPort + "/mqtt",
+                        mqttClientId,
+                        new MemoryPersistence());
+            }
+
+            customWebsocketHeaders = new Properties();
```
jamesonwilliams: Where is this declared? Can it be local to this method? Or alternately, can you instantiate the `new Properties()` closer to the declaration (Constructor?)

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
desokroshan: That makes lot of sense. It will definitely save us a lot of NPEs. Updated to follow the pattern.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -941,6 +947,72 @@ public void run() {
...
+    public void connect(
+            final String tokenKeyName,
+            final String token,
+            final String tokenSignature,
+            final String customAuthorizer,
+            final AWSIotMqttClientStatusCallback statusCallback) {
+
+        if (tokenKeyName == null || token == null || tokenSignature == null) {
+            throw new IllegalArgumentException("tokenKeyName/token/tokenSignature cannot be null");
+        }
```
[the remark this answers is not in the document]
desokroshan: This call seems to require API Level of 19 while the current minimum for IoT SDK is 11. We should probably increase the minimum API level but I feel 19 would too high given our user demographics.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
desokroshan: This is an instance variable because it is reused for reconnection. I have moved the instantiation to Constructor anyways.

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -724,37 +724,39 @@ public void onFailure(Throwable exception) {
...
 
-        if (authMode.equals(CUSTOM_AUTH)) {
+        if (AuthenticationMode.CUSTOM_AUTH.equals(authMode)) {
             // connect using WebSockets and custom authentication token
             mqttManager.connect(
-                    "token-header",
-                    "allow",
-                    "Gv85h71GT07RWo5uWQpqHlYWePN7YVY7KgiuPVEnt65+tDDWyZOlGLIDxvwC7LiBF7FbmG/coeBELKltqmrP0gjM7sIyXlrx315xG3LupoK4bmsNIzg6xxWa0n60WxhOv1p5qXDAhmYWIFuL0enMtuwpCY1W073NqZ/+iOsrJexssZiiLnvVf3z6HReEEER6TfFevrgAEUva0h9pzbUepbJ6RaOO7zZY6e/T/wSHrKl0jA4MalGReKbhdlg0HyTPfILhjorYkF7ew/8Ml+QMr/WU90YMp9Sv8SF0q7WBzbuqxpAR8vxm00rRlgrFdsIRjGFTDmrp9Fsx+xlx0aADCA==",
-                    "CustomAuthorizerIntegTest",
+                    getPackageConfigure().getString("token_key_name"),
```
jamesonwilliams: Most ideal would be to get this config object into a statically typed object model. That is, something like:
```
final class ConnectionOptions {
    // Can some of these things be enumerations / validated / constrained beyond just String?
    private final String tokenKeyName;
    private final String token;
    private final String tokenSignature;
    private final String customerAuthorizerName;

    ...
    static ConnectionOptions fromPackageConfig(PackageConfig config) {
        return new ConnectionOptions(
            config.getString(ConfigKey.TOKEN_KEY_NAME),
            config.getString(ConfigKey.TOKEN),
            config.getString(ConfigKey.TOKEN_SIGNATURE),
            config.getString(ConfigKey.CUSTOMER_AUTHORIZER_NAME)
        );
    }
}

enum ConfigKey {
    TOKEN_KEY_NAME("token_key_name"),
    TOKEN("token"),
    TOKEN_SIGNATURE("token_signature"),
    CUSTOM_AUTHORIZER_NAME("custom_authorizer_name")

    private final String value;

    @Override
    public String toString() {
        return value;
    }

    public static ConfigKey formString(@Nullable String value) {
        for (ConfigKey value : values()) {
            if (value.toString().equals(value)) {
               return value;
            }
        }
        throw new IllegalArgumentException("No config key with value = " + value);
    }
}
```
    


```

File: aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java [hunk 8]
```diff
@@ -804,37 +806,39 @@ public void onMessageArrived(String topic, byte[] data) {
...
             });
-        } else if (authMode.equals(CUSTOM_AUTH)) {
+        } else if (AuthenticationMode.CUSTOM_AUTH.equals(authMode)) {
             // connect using WebSockets and custom authentication token
             mqttManager.connect(
-                    "token-header",
-                    "allow",
-                    "Gv85h71GT07RWo5uWQpqHlYWePN7YVY7KgiuPVEnt65+tDDWyZOlGLIDxvwC7LiBF7FbmG/coeBELKltqmrP0gjM7sIyXlrx315xG3LupoK4bmsNIzg6xxWa0n60WxhOv1p5qXDAhmYWIFuL0enMtuwpCY1W073NqZ/+iOsrJexssZiiLnvVf3z6HReEEER6TfFevrgAEUva0h9pzbUepbJ6RaOO7zZY6e/T/wSHrKl0jA4MalGReKbhdlg0HyTPfILhjorYkF7ew/8Ml+QMr/WU90YMp9Sv8SF0q7WBzbuqxpAR8vxm00rRlgrFdsIRjGFTDmrp9Fsx+xlx0aADCA==",
-                    "CustomAuthorizerIntegTest",
+                    getPackageConfigure().getString("token_key_name"),
```
jamesonwilliams: Same here, I'd at least pull out the `getFoo()` call so you just have `config.getString("key_blah")`. For "extra credit" (lol) you could do the "strong modeling" of the keys and options object, too.

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
desokroshan: I have transformed the connection params into a statically typed object. Quick note : None of the connection parameters viz tokenKeyName, token, tokenSignature and CustomAuthorizerName can be constrained in any way since these could be any string determined by the user.

desokroshan:
> Cool, looks reasonable - do we have a plan yet for how this gets tested - unit, component, integration, etc.?

The integration test I have added validates the following scenarios : 
1. Connection with custom authentication token
2. Reconnection with custom authentication token if auto-reconnect is set to true
3. Publish to an IoT topic
4. Subscription to a topic and to ensure that messages published on subscribed topic are received by the subscriber.

Besides the integration tests, I have validated these scenarios through the sample app.

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -722,21 +727,48 @@ public void onFailure(Throwable exception) {
...
         mqttManager.setAutoReconnect(false);
-        // connect using WebSockets and IAM credentials
-        mqttManager.connect(credentialsProvider, new AWSIotMqttClientStatusCallback() {
-            @Override
-            public void onStatusChanged(AWSIotMqttClientStatus status, Throwable throwable) {
-                statuses.add(status);
-            }
-        });
+
+        if (AuthenticationMode.CUSTOM_AUTH.equals(authMode)) {
```
jamesonwilliams:
```
switch (authMode) {
    case CUSTOM_AUTH:
}
```

In the `if`, you use just `AuthenticationMode`, but in the `else if`, you use `AWSIotMqttManager.AuthenticationMode`. I'd pick one and stick with it. So long as you can use the second one without getting any very long lines, I'd probably use it.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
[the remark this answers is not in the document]
jamesonwilliams: Ah, true. We can use it back to 15 in Amplify cause it gets "de-sugared". Anyway, sounds like its not worth the effort here since we use an old Java version and don't use retrolambda.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -941,6 +949,71 @@ public void run() {
...
+
+        this.userStatusCallback = statusCallback;
+
+        // Do nothing if Connecting, Connected or Reconnecting
+        if (connectionState != MqttManagerConnectionState.Disconnected) {
```
jamesonwilliams: Change this to `NonNull.equals(possibleNull)`. See if there are any more of these in this PR, remaining.

desokroshan: > Are the SDK unit tests supposed to be passing, at this point?

Yep. I suppose there was a transient failure in the last circleci build. It all passes in the latest build.

desokroshan:
> Are the SDK unit tests supposed to be passing, at this point?

Yep. I suppose there was a transient failure in the last circleci build. It all passes in the latest build.

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
desokroshan: Updated with `AWSIotMqttManager.AuthenticationMode`. I agree that this is more descriptive and preferable.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
desokroshan: I came across an interesting SO post on whether to use the operator `==` or `equals()` method for enum comparison and it looks like the [operator is preferred](https://stackoverflow.com/questions/1750435/comparing-java-enum-members-or-equals) because of null-safety and other reasons but they are the same for all practical purposes. I have updated the occurrences in PR with `NonNull.equals(possibleNull)` pattern.

PR #1509, opened 2020-03-09
[Infrastructure] Release latest javadoc after release to maven is complete

desokroshan:
*Issue #, if available:*

*Description of changes:*
Currently, javadocs are released in parallel with the release of binaries to maven. If release to maven fails javadocs talk of the latest version which would not be publicly available. This PR ensures that the javadocs are released after the maven release is complete.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1957,7 +1957,6 @@ workflows:
       - release_s3:
           requires:
             - build
-            - release_javadoc
             - release_maven
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1948,6 +1948,7 @@ workflows:
           requires:
             - unittest
             - post_integrationtest
+            - release_maven
           filters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1948,6 +1948,7 @@ workflows:
           requires:
             - unittest
             - post_integrationtest
+            - release_maven
```
raphkim: is this not supposed to be `release_javadoc`?

desokroshan: What I intend here is for release_javadoc to depend on release_maven. So `release_javadoc` job should have `release_maven` in its `requires` directive. This ensures that javadoc is released after maven is released.

raphkim: looks like `release_javadoc` isn't anywhere anymore?

raphkim: ah I kind of see it. I guess I was just confused from not seeing the rest of context. LGTM

PR #1519, opened 2020-03-17
HttpUrlClient improvements to make more extendible

lgoldstein:
*Issue #, if available:*

*Description of changes:*

* (most important) use the already existing `getUrlConnection` method to obtain the connection from the request URL. 

* (useful) promote visibility of private and package scope methods and fields to `protected` - after all, the idea of an SDK is to allow users to extend the functionality

The rationale behind this is twofold:

* use the already existing method - which was not used even though it gave the impression that it was.

* provide a way for users to add customized settings to the connection that currently we cannot foresee - e.g., add headers, provide their own trust manager - etc...

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/http/UrlHttpClient.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/http/UrlHttpClient.java
@@ -66,7 +66,7 @@ public class UrlHttpClient implements HttpClient {
     @Override
     public HttpResponse execute(final HttpRequest request) throws IOException {
         final URL url = request.getUri().toURL();
-        final HttpURLConnection connection = (HttpURLConnection) url.openConnection();
+        final HttpURLConnection connection = getUrlConnection(url);
         final CurlBuilder curlBuilder = config.isCurlLogging()
                 ? new CurlBuilder(request.getUri().toURL()) : null;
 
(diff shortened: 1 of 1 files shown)


jamesonwilliams:
@lgoldstein Thanks for this suggestion. From your description, I understand that you want to extend/change/override the default behavior of the `UrlHttpClient` through _inheritence_.

So that we can better understand why, could you provide an example subclass of `UrlHttpClient`, which contains the custom functionality you are trying to achieve? Inheritance may not be the only mechanism to enable the use case(s).

Thanks!

lgoldstein:
>> could you provide an example subclass of UrlHttpClient, which contains the custom functionality you are trying to achieve? 

I was interested primarily in `UrlHttpClient#getUrlConnection ` (which already is _protected_ but never invoked). Without going into too many details, I wanted to know the exact URL being used in order to dynamically enable a  "hole" through a firewall based on the outgoing request details.

Later on, I needed to add some more headers and perhaps configure the URL connection timeouts or the SSL trust manager  based on the specific request being made - something that the other methods had they been _protected_ as well would have allowed me to do.

>> Inheritance may not be the only mechanism to enable the use case(s).

Agreed, but as an open source project maintainer myself I believe that the end users may have needs that might not have been foreseen at the time the code was written, so if we can facilitate it by something as simple as allowing them to override a method then where's the harm ? I agree that inheritance might not be the answer, but IMO the **users** are the ones that are using the code and thus best qualified to decide whether the mechanism is adequate or not - and they are the ones who will have to abide by that decision.

jamesonwilliams:
We would prefer to leave the `UrlHttpClient` class alone. The `HttpClient` is intended for extension. You could create a new implementation of that interface for your needs.

lgoldstein:
>>  The HttpClient is intended for extension. You could create a new implementation of that interface for your needs.

That's what I meant by the suggested changes - instead of re-implementing `HttpClient` I was going to override 1-2 methods of  `UrlHttpClient`. Isn't this what open source is all about ? Let the users decide how best to use the library and facilitate it ?

PR #1520, opened 2020-03-17
Bug Fixed: Added exception handling in Sign In options

ShubhamsGeekyAnts:
signIn and showSignIn method can throw Exception when user is already signed in


raphkim:
Hi @ShubhamsGeekyAnts ,

Thank you for your contribution to `AWSMobileClient`. I will review review your suggestion with team and accept the changes once we confirm the expected behavior.

jamesonwilliams:
Closing per @raphkim's suggestion to user `isSignedIn()` _outside_the API surface.

ShubhamsGeekyAnts:
Hi, @raphkim 
If user is already signed in and try to send sign in request again, then SDK should through an exception. I took idea of `isSignedIn()` check before sign in request from iOS SDK for `AWSMobileClient `. It throws an exception when any session is already found.

Issue #1469, opened 2020-02-13
AWSMobileCLient getTokens returns Exception: No cached session.


TobiS97:
**State your question**
I followed the Tutorial on https://aws-amplify.github.io/docs/sdk/android/authentication. After a sucessfully calling the signIn Method from the AWSMobileCLient as stated in the tutorial, I get back the message "Tokens are invalid, please sign-in again. Exception: No chached session." 
The userState of my user is logged in.
The essential code:

        AWSMobileClient.getInstance().signIn(username, password, null, new Callback<SignInResult>               () {
            @Override
            public void onResult(final SignInResult signInResult) {
                runOnUiThread(new Runnable() {
                    @Override
                    public void run() {
                        Log.d("CognitoService", "Sign-in callback state: " + signInResult
                                .getSignInState());
                        switch (signInResult.getSignInState()) {
                            case DONE:
                                makeToast(context, "Sign-in done.");
                                confirmSignIn();
                                try {
                                    Log.d("Debug", AWSMobileClient.getInstance().getTokens().getAccessToken
                                            ().toString());
                                } catch (Exception e) {
                                    e.printStackTrace();
                                }
                                Intent intent = new Intent(getApplicationContext(), MainActivity
                                        .class);
                                startActivity(intent);
                                break;
                            case SMS_MFA:
                                makeToast(context, "Please confirm sign-in with SMS.");
                                break;
                            case NEW_PASSWORD_REQUIRED:
                                makeToast(context, "Please confirm sign-in with new password.");
                                break;
                            default:
                                makeToast(context, "Unsupported sign-in confirmation: "
                                        + signInResult.getSignInState());
                                break;
                        }
                    }
                });
            }`
Here the stacktrace of the exception:
It is weird that the getTokens() call delegates the logic to a _getHostedUITokens() method, even if I don't use the hosted ui in this setup. I only use the hosted UI when clicking a login with FB button.

```
02-13 10:44:13.647 28585-28585/com.example.tobi.androidapp W/AWSMobileClient: Tokens are invalid, please sign-in again.
java.lang.Exception: No cached session.
    at com.amazonaws.mobile.client.AWSMobileClient$12.onFailure(AWSMobileClient.java:1818)
    at com.amazonaws.mobileconnectors.cognitoauth.AuthClient.getSession(AuthClient.java:182)
    at com.amazonaws.mobileconnectors.cognitoauth.Auth.getSession(Auth.java:674)
[...]

Issue #1262, opened 2019-10-19
I get "No cached session error" after calling getToken()  when user state is SIGNED_IN 


mohsenk:
**Describe the bug**
At first when i login with Google HosteUI everything is good. but after close and open app. my userStateListener  says that user is "SIGNED_IN" and i'm going to get call getTokens() method but it returns error "No Cached Session".

```
2019-10-19 14:38:49.108 12558-12660/hu.avicenna.android E/ApiClient: error in get token
    java.lang.Exception: No cached session.
        at com.amazonaws.mobile.client.AWSMobileClient$10.onFailure(AWSMobileClient.java:1648)
        at com.amazonaws.mobileconnectors.cognitoauth.AuthClient.getSession(AuthClient.java:182)
        at com.amazonaws.mobileconnectors.cognitoauth.Auth.getSession(Auth.java:674)
        at com.amazonaws.mobile.client.AWSMobileClient._getHostedUITokens(AWSMobileClient.java:1651)
        at com.amazonaws.mobile.client.AWSMobileClient.access$800(AWSMobileClient.java:163)
        at com.amazonaws.mobile.client.AWSMobileClient$9.run(AWSMobileClient.java:1573)
        at com.amazonaws.mobile.client.internal.InternalCallback$1.run(InternalCallback.java:101)
        at java.lang.Thread.run(Thread.java:764)
     Caused by: java.lang.Exception: No cached session
        at com.amazonaws.mobileconnectors.cognitoauth.AuthClient.getSession(AuthClient.java:182) 
        at com.amazonaws.mobileconnectors.cognitoauth.Auth.getSession(Auth.java:674) 
        at com.amazonaws.mobile.client.AWSMobileClient._getHostedUITokens(AWSMobileClient.java:1651) 
        at com.amazonaws.mobile.client.AWSMobileClient.access$800(AWSMobileClient.java:163) 
        at com.amazonaws.mobile.client.AWSMobileClient$9.run(AWSMobileClient.java:1573) 
        at com.amazonaws.mobile.client.internal.InternalCallback$1.run(InternalCallback.java:101) 
        at java.lang.Thread.run(Thread.java:764) 
```

**Environment Information (please complete the following information):**
 - AWS Android SDK Version: 2.15.1
 - Device: OnePlus 5T
 - Android Version: 8.1
 - Specific to simulators: No

kvasukib:
[...]

PR #1531, opened 2020-03-25
Fixes sign in method caching issues #1469 and #1262

TrekSoft:
Currently SIGN_IN_MODE is only set when the user signs in with HostedUI or OAuth leading to a failure in retrieving tokens if the user afterwards ever tries to sign in with UserPools or Federated Identities. 

This ensures that the sign in mode is switched when those methods are used.


TrekSoft:
Thanks @undefobj - I thought about that but at this point don't have the time commitment to write that so feel like it's better to at least stop the bleeding for now and also we're hoping to replace AWSMobileClient anyway

karmaobserver:
Hello @TrekSoft, it passed a few days from release 2.16.10 version and there are change logs but the library is not uploaded on maven repo.

TrekSoft:
Hey @karmaobserver - sorry for that - investigating how it missed getting uploaded to maven.

karmaobserver:
Thank you very much, it's up now.

Issue #1500, opened 2020-03-03
pause in TransferUtility seems not work


gcobc12677:
**Describe the bug**
with code below,  value returned by pause is true,
but `onProgressChanged` in the listener keeps running

in version: 2.11.0, though state `FAILED` comes up after `PAUSED` in `onStateChanged`,
uploading task stop properly.

**To Reproduce**
```
TransferObserver uploadObserver = getS3Utility(context).upload(
        //...
);
uploadObserver.setTransferListener(getmTransferListener(this));
```

start an upload via above method, then pause by its id like

```
getS3Utility(aContext).pause(taskId); // taskId is equal to uploadObserver.getId()
```

**Which AWS service(s) are affected?**
com.amazonaws:aws-android-sdk-s3:2.16.8
com.amazonaws:aws-android-sdk-mobile-client:2.16.8

**Expected behavior**
when pause method is triggered, uploading task stop

**Environment Information (please complete the following information):**
 - AWS Android SDK Version:2.16.8
 - Device: Pixel 3
 - Android Version: 10
 - Specific to simulators: No

raphkim:
Hi @gcobc12677 ,

Thank you for taking time to report the issue. To allow us to assist you better, would you be able to provide the relevant code snippet where you are setting up the `TransferUtility` as well as the log at the time of transfer failure?

gcobc12677:
below is how I get instances of `AmazonS3` and `TransferUtility`

```
private static AmazonS3 s3Client = null;
private static TransferUtility s3Utility = null;

private static AmazonS3 getS3Client() {
    if (s3Client == null) {
        BasicAWSCredentials credentials = new BasicAWSCredentials(ACCESS_KEY, SECRET_KEY);

        s3Client = Observable.create(
                (ObservableOnSubscribe<AmazonS3>) emitter -> {
                    s3Client = new AmazonS3Client(
                            credentials,
                            Region.getRegion(Regions.AP_NORTHEAST_1)
                    );
                    emitter.onNext(s3Client);
                    emitter.onComplete();
                })
                .singleElement()
                .blockingGet();
    }
    return s3Client;
}

private static TransferUtility getS3Utility(Context applicationContext) {
    if (s3Utility == null) {
        TransferNetworkLossHandler.getInstance(applicationContext);
        s3Utility = TransferUtility
                .builder()
                .context(applicationContext)
                .s3Client(getS3Client())
                .build();
    }

    return s3Utility;
}
```

[...]

PR #1536, opened 2020-03-29
Fix multipart upload pause

raphkim:
*Issue #, if available:*
#1500 

*Description of changes:*
`TransferUtility#pause(int)` was not behaving correctly when uploading a multipart (> 5MB) item. This was due to interruption detection (due to pause) code returning early before actually canceling the individual `UploadPartTask` tasks in the upload. The same result would have been seen for `TransferUtility#cancel(int)` for a large upload.

Additionally to fixing the above bug, this PR also contains fix to mitigate the effects of race condition observed when pausing multipart upload and resuming it instantly. Previously, the transfer state would be set to `PAUSED` or `CANCELED` as soon as the method was called even before all of the task was interrupted. Now `PENDING_PAUSE` and `PENDING_CANCEL` (already present in the codebase but unused) states are used to make sure that the task is successfully interrupted before the state is updated to its final state.

### DIFF
--- a/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/mobileconnectors/s3/transferutility/PauseTransferIntegrationTest.java
+++ b/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/mobileconnectors/s3/transferutility/PauseTransferIntegrationTest.java
@@ -1,0 +1,156 @@
+/**
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ *  http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazonaws.mobileconnectors.s3.transferutility;
+
+import android.content.Context;
+import android.support.test.InstrumentationRegistry;
+
+import com.amazonaws.services.s3.S3IntegrationTestBase;
+
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.io.File;
+import java.util.Date;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public final class PauseTransferIntegrationTest extends S3IntegrationTestBase {
+
+    private static final String bucketName = "amazon-transfer-util-integ-test-" + new Date().getTime();
+    private static TransferUtility util;
+    private static Context context = InstrumentationRegistry.getContext();
+
+    private File file;
+    private CountDownLatch started;
+    private CountDownLatch paused;
+    private CountDownLatch resumed;
+    private CountDownLatch completed;
+
+    /**
+     * Creates and initializes all the test resources needed for these tests.
+     */
+    @BeforeClass
+    public static void setUpBeforeClass() throws Exception {
+        setUp();
+        TransferNetworkLossHandler.getInstance(context);
+        util = TransferUtility.builder()
+                .context(context)
+                .s3Client(s3)
+                .build();
+
+        try {
+            s3.createBucket(bucketName);
+            waitForBucketCreation(bucketName);
+        } catch (final Exception e) {
+            System.out.println("Error in creating the bucket. "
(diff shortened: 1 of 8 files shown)


aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/mobileconnectors/s3/transferutility/PauseTransferIntegrationTest.java
```diff
@@ -0,0 +1,156 @@
...
+import org.junit.Test;
+
+import java.io.File;
+import java.util.Date;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+
+public final class PauseTransferIntegrationTest extends S3IntegrationTestBase {
+
+    private static final String bucketName = "amazon-transfer-util-integ-test-" + new Date().getTime();
```
desokroshan: Can we name it like `BUCKET_NAME` to be consistent with convention and rest of the codebase?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/DownloadTask.java
```diff
@@ -144,8 +144,6 @@ public Boolean call() {
                      */
                     updater.updateState(download.id, TransferState.WAITING_FOR_NETWORK);
                     LOGGER.debug("Network Connection Interrupted: " + "Moving the TransferState to WAITING_FOR_NETWORK");
-                    ProgressEvent resetEvent = new ProgressEvent(0);
-                    resetEvent.setEventCode(ProgressEvent.RESET_EVENT_CODE);
```
desokroshan: Shouldn't we be resetting the progress upon loss of network and then pass `reserEvent` to the `progressListener`?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadTask.java
```diff
@@ -310,8 +315,6 @@ private Boolean uploadSinglePartAndWaitForCompletion() {
                      */
                     updater.updateState(upload.id, TransferState.WAITING_FOR_NETWORK);
                     LOGGER.debug("Network Connection Interrupted: " + "Moving the TransferState to WAITING_FOR_NETWORK");
-                    ProgressEvent resetEvent = new ProgressEvent(0);
-                    resetEvent.setEventCode(ProgressEvent.RESET_EVENT_CODE);
```
desokroshan: Same as above. If `eventCode` is not being used anywhere should we consider removing it altogether from the `ProgressEvent` class?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/DownloadTask.java
```diff
@@ -114,16 +114,16 @@ public Boolean call() {
...
+                updater.updateState(download.id, TransferState.CANCELED);
+                LOGGER.info("Transfer is " + TransferState.CANCELED);
                 return false;
             }
 
             // Reset the progress when the transfer is paused.
-            if (TransferState.PAUSED.equals(download.state)) {
-                LOGGER.info("Transfer is " + download.state);
-                ProgressEvent resetEvent = new ProgressEvent(0);
-                resetEvent.setEventCode(ProgressEvent.RESET_EVENT_CODE);
```
desokroshan: Are we not resetting on pause on anymore? If so, do we need to update documentation?

raphkim: if you look closely, the lines i removed were noops. Im not sure if the original intent was to pass in a `new ProgressEvent(0)` with reset flag, but this object never gets used and a new progress event without flag is reinstantiated instead. Perhaps a different bug, but I didn't want to introduce yet another behavior change in this PR.

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/mobileconnectors/s3/transferutility/PauseTransferIntegrationTest.java
raphkim: yea oops. A mistake that was copy + pasted from my previous mistake. will fix them

raphkim:
This PR contained removal of some lines that SEEMED to be doing something important but was actually not doing anything:

```
ProgressEvent resetEvent = new ProgressEvent(0);
resetEvent.setEventCode(ProgressEvent.RESET_EVENT_CODE);
progressListener.progressChanged(new ProgressEvent(0));
```

The first two lines don't do anything, since `resetEvent` is not used anywhere after this bit of code. 

This is probably an important issue of its own that requires its own investigation, so I decided to leave the redundant lines for now to revisit in the future with a separate PR.

PR #1538, opened 2020-04-01
[s3] Remove unnecessary public access to S3 buckets created under test

desokroshan:
*Issue #, if available:*

*Description of changes:*


aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/CleanupBucketIntegrationTests.java
```diff
@@ -61,7 +61,11 @@ public void testCleanup() {
                     || bucket.getName().startsWith("s3-low-level-presigned-url")
                     || bucket.getName().startsWith("java-bucket-policy-integ-test")
                     || bucket.getName().startsWith("copy-object-integ-test")
-                    || bucket.getName().startsWith("java-server-side-encryption-integ-test")) {
+                    || bucket.getName().startsWith("java-server-side-encryption-integ-test")
+                    || bucket.getName().startsWith("acceleration-integration-test-")
+                    || bucket.getName().startsWith("amazon-transfer-util-integ-test")
+                    || bucket.getName().startsWith("java-bucket-lifecycle-integ-test-")
+                    || bucket.getName().startsWith("java-s3-object-iteration-test-")) {
```
jpignata, line 68: It's hard to look at this without thinking there's a better way to approach this such as a standard prefix or using S3 tags to indicate it's a test bucket. Thoughts?

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/UnauthenticatedIntegrationTest.java
```diff
@@ -65,7 +65,6 @@ public void tearDown() {
     @Test
     public void testUnauthenticatedOperations() {
         s3.createBucket(bucketName);
-        s3.setBucketAcl(bucketName, CannedAccessControlList.PublicRead);
```
jpignata, line 68: Is this change sufficient? How will the below line that calls ` PutObjectRequest` succeed? Did you run these tests?

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/V1BucketAddressingIntegrationTest.java
```diff
@@ -75,7 +75,6 @@ public void testV1BucketAddressing() throws Exception {
         createV1AddressedBucket(bucketName);
         file = super.getRandomTempFile("v1-addressing-integ-test", 333L);
 
-        s3.setBucketAcl(bucketName, CannedAccessControlList.PublicRead);
```
jpignata, line 78: Same question. What is this testing, and why was this public?

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/UnauthenticatedIntegrationTest.java
jpignata: So essentially this line was unnecessary and the test still passes without it?

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/V1BucketAddressingIntegrationTest.java
jpignata: Same here?

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/UnauthenticatedIntegrationTest.java
desokroshan: Yeah these seem redundant and the tests still pass without it, so removing would suffice in this case.

desokroshan:
> What's the plan for the other reference(s)?
> 
> For example:
> 
> https://github.com/aws-amplify/aws-sdk-android/blob/881bb946777a56488c659c8ef8c210d1bf689f3d/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/AclIntegrationTest.java#L136

Here the test explicitly is to check if we can set public read ACL on a bucket and there is a assert statement to this effect. So I don't think there is a way to remove this without actually getting rid of the test case itself.

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/CleanupBucketIntegrationTests.java
desokroshan: We can for some of these but interestingly a lot these patterns are not created by our integration tests. I guess these are being created by tests from other teams (not sure why?). So we might have trace their origin and make a fix accordingly.

jpignata:
> Here the test explicitly is to check if we can set public read ACL on a bucket and there is a assert statement to this effect. So I don't think there is a way to remove this without actually getting rid of the test case itself.

The idea here is to test our code and its ability to set a canned ACL -- not for us to test every canned ACL. Why don't we simply remove this canned ACL and test others that don't potentially introduce security concerns? What are the downsides in terms of the code we're trying to exercise?

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/CleanupBucketIntegrationTests.java
jpignata: We're sharing the account with other processes? If so, let's fix that immediately to make this easier.

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/CleanupBucketIntegrationTests.java
```diff
@@ -61,7 +61,11 @@ public void testCleanup() {
                     || bucket.getName().startsWith("s3-low-level-presigned-url")
                     || bucket.getName().startsWith("java-bucket-policy-integ-test")
                     || bucket.getName().startsWith("copy-object-integ-test")
-                    || bucket.getName().startsWith("java-server-side-encryption-integ-test")) {
+                    || bucket.getName().startsWith("java-server-side-encryption-integ-test")
+                    || bucket.getName().startsWith("acceleration-integration-test-")
```
jamesonwilliams, line 65: The existing code construct here is absolutely batty.

At _minimum_, it should be like:
```
List<String> bucketPrefixes = Arrays.asList(
    "acceleration-integration-test-",
   ...
);
for (String prefix : bucketPrefixes) {
    if (bucket.getName().startsWith(prefix)) {
        // do stuff
    }
}
```

From there, generating the `List<String>` in a more automated way (from file, from inspection of test data, etc.) would be desirable.

@jpignata's suggestion of using an AWS Tag on the bucket like `special-test-slug-55` or whatever would also be a nice improvement.

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/mobileconnectors/s3/transferutility/PauseTransferIntegrationTest.java
```diff
@@ -32,7 +32,7 @@
 
 public final class PauseTransferIntegrationTest extends S3IntegrationTestBase {
 
-    private static final String BUCKET_NAME = "amazon-transfer-util-integ-test-" + new Date().getTime();
+    private static final String BUCKET_NAME = "android-sdk-transfer-util-integ-test-" + new Date().getTime();
```
jpignata, line 35: In some of these constructions, we're using `Date().getTime()`, in others: `System.currentTimeMillis()` -- what do you think about being consistent?

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/CleanupBucketIntegrationTests.java
desokroshan: Updated to use a common prefix for all the test buckets.

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/AclIntegrationTest.java
```diff
@@ -144,8 +143,6 @@ public void testCannedAcls() throws Exception {
         for (int index = 0; index < accessControls.length; index++) {
```
jpignata, line 143: What is this test enforcing now? That only assert is checking bucket owner, which seems far removed from its original intent.

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/AclIntegrationTest.java
```diff
@@ -124,35 +125,35 @@ public void testCreateBucketWithAcl() throws Exception {
...
         // Public Read Canned ACL
-        s3.setBucketAcl(bucketName, CannedAccessControlList.PublicRead);
-        s3.setObjectAcl(bucketName, key, CannedAccessControlList.PublicRead);
-        AccessControlList bucketAcl = s3.getBucketAcl(bucketName);
-        AccessControlList objectAcl = s3.getObjectAcl(bucketName, key);
+        s3.setObjectAcl(BUCKET_NAME, KEY, CannedAccessControlList.PublicRead);
+        AccessControlList bucketAcl = s3.getBucketAcl(BUCKET_NAME);
+        AccessControlList objectAcl = s3.getObjectAcl(BUCKET_NAME, KEY);
 
         AccessControlList[] accessControls = new AccessControlList[] {
```
jpignata, line 141: What is this used for? I see the variable being overridden below, is this necessary, or should we just move the declration to line 157?

PR #1542, opened 2020-04-03
Rjuliano/journeys

rjuliano:
*Issue #, if available:*

*Description of changes:*
This PR adds support for push notifications from Pinpoint Journeys.

There are several commits. With the exception of the most recent, they mostly have to do with renaming and refactoring parts of the code that treated every push notification as being part of a campaign. I've changed that to be `eventSource` where appropriate. So any time you see eventSource throughout the code, think Pinpoint Campaign or Journey (or some other future thing)

The latest commit is the one that actually deals with adding an API (`handlePushNotification`) that will be able to handle both Campaigns and Journeys.

### DIFF
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
@@ -75,7 +75,7 @@ public class AnalyticsClient implements JSONSerializable {
     private final Map<String, Double> globalMetrics = new ConcurrentHashMap<String, Double>();
     private final Map<String, Map<String, String>> eventTypeAttributes = new ConcurrentHashMap<String, Map<String, String>>();
     private final Map<String, Map<String, Double>> eventTypeMetrics = new ConcurrentHashMap<String, Map<String, Double>>();
-    private Map<String, String> eventSourceAttributes = new ConcurrentHashMap<String, String>();
+    private final Map<String, String> eventSourceAttributes = new ConcurrentHashMap<String, String>();
     private String sessionId;
     private long sessionStartTime;
     private EventRecorder eventRecorder;
@@ -359,6 +359,16 @@ public class AnalyticsClient implements JSONSerializable {
     }
 
     /**
+     *
+     * @param campaign
+     * @deprecated see {@link #updateEventSourceGlobally(Map)}
+     */
+    @Deprecated
+    public void setCampaignAttributes(Map<String, String> campaign) {
+        updateEventSourceGlobally(campaign);
+    }
+
+    /**
      * Replaces the current event source attributes (if any)
      * with a new set of event source attributes
      * @param attributes map of attribute values
@@ -374,7 +384,7 @@ public class AnalyticsClient implements JSONSerializable {
         }
         //We set this so the analytics client knows what to remove from
         //from the globalAttributes collection when it's time to clear it
-        this.eventSourceAttributes = attributes;
+        eventSourceAttributes.putAll(attributes);
     }
 
     /**
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
@@ -75,7 +75,7 @@ public class AnalyticsClient implements JSONSerializable {
     private final Map<String, Double> globalMetrics = new ConcurrentHashMap<String, Double>();
     private final Map<String, Map<String, String>> eventTypeAttributes = new ConcurrentHashMap<String, Map<String, String>>();
     private final Map<String, Map<String, Double>> eventTypeMetrics = new ConcurrentHashMap<String, Map<String, Double>>();
-    private Map<String, String> campaignAttributes = new ConcurrentHashMap<String, String>();
+    private Map<String, String> pinpointAttributes = new ConcurrentHashMap<String, String>();
     private String sessionId;
     private long sessionStartTime;
     private EventRecorder eventRecorder;
@@ -359,33 +359,33 @@ public class AnalyticsClient implements JSONSerializable {
     }
 
     /**
-     * Adds the specified campaign attributes to events to track Campaign Analytic
+     * Adds the specified pinpoint attributes to events to track Campaign/Journey Analytic
      * <p>
      * You should not use this method as it will be called by the NotificationManager when the app is opened
      * from a push notification.
      *
-     * @param campaign the map with campaign attributes of the campaign received
+     * @param pinpointAttributes the map with pinpointAttributes attributes of the pinpointAttributes received
      */
-    public void setCampaignAttributes(Map<String, String> campaign) {
-        if (campaign == null) {
-            log.warn("Null campaign attributes provided to setCampaignAttributes.");
+    public void setPinpointAttributes(Map<String, String> pinpointAttributes) {
+        if (pinpointAttributes == null) {
+            log.warn("Null pinpointAttributes attributes provided to setPinpointAttributes.");
             return;
         }
 
-        campaignAttributes = campaign;
+        this.pinpointAttributes = pinpointAttributes;
     }
 
     /**
-     * Clears campaign attributes
+     * Clears pinpoint attributes
      * <p>
      * You should not use this method as it will be called by the NotificationManager when the app is opened
      * from a push notification.
      */
-    public void clearCampaignAttributes() {
-        for (final String key : campaignAttributes.keySet()) {
+    public void clearPinpointAttributes() {
+        for (final String key : pinpointAttributes.keySet()) {
             this.removeGlobalAttribute(key);
         }
-        campaignAttributes.clear();
+        pinpointAttributes.clear();
     }
 
     @Override
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
@@ -359,7 +359,7 @@ public class AnalyticsClient implements JSONSerializable {
     }
 
     /**
-     * Adds the specified pinpoint attributes to events to track Campaign/Journey Analytic
+     * Adds the specified Pinpoint attributes to events to track Campaign/Journey analytics
      * <p>
      * You should not use this method as it will be called by the NotificationManager when the app is opened
      * from a push notification.
@@ -376,7 +376,7 @@ public class AnalyticsClient implements JSONSerializable {
     }
 
     /**
-     * Clears pinpoint attributes
+     * Clears Pinpoint attributes
      * <p>
      * You should not use this method as it will be called by the NotificationManager when the app is opened
      * from a push notification.
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
@@ -75,7 +75,7 @@ public class AnalyticsClient implements JSONSerializable {
     private final Map<String, Double> globalMetrics = new ConcurrentHashMap<String, Double>();
     private final Map<String, Map<String, String>> eventTypeAttributes = new ConcurrentHashMap<String, Map<String, String>>();
     private final Map<String, Map<String, Double>> eventTypeMetrics = new ConcurrentHashMap<String, Map<String, Double>>();
-    private Map<String, String> pinpointAttributes = new ConcurrentHashMap<String, String>();
+    private Map<String, String> eventSourceAttributes = new ConcurrentHashMap<String, String>();
     private String sessionId;
     private long sessionStartTime;
     private EventRecorder eventRecorder;
@@ -364,15 +364,15 @@ public class AnalyticsClient implements JSONSerializable {
      * You should not use this method as it will be called by the NotificationManager when the app is opened
      * from a push notification.
      *
-     * @param pinpointAttributes the map with pinpointAttributes attributes of the pinpointAttributes received
+     * @param eventSourceAttributes the map with pinpointAttributes attributes of the pinpointAttributes received
      */
-    public void setPinpointAttributes(Map<String, String> pinpointAttributes) {
-        if (pinpointAttributes == null) {
+    public void setEventSourceAttributes(Map<String, String> eventSourceAttributes) {
+        if (eventSourceAttributes == null) {
             log.warn("Null pinpointAttributes attributes provided to setPinpointAttributes.");
             return;
         }
 
-        this.pinpointAttributes = pinpointAttributes;
+        this.eventSourceAttributes = eventSourceAttributes;
     }
 
     /**
@@ -381,11 +381,11 @@ public class AnalyticsClient implements JSONSerializable {
      * You should not use this method as it will be called by the NotificationManager when the app is opened
      * from a push notification.
      */
-    public void clearPinpointAttributes() {
-        for (final String key : pinpointAttributes.keySet()) {
+    public void clearEventSourceAttributes() {
+        for (final String key : eventSourceAttributes.keySet()) {
             this.removeGlobalAttribute(key);
         }
-        pinpointAttributes.clear();
+        eventSourceAttributes.clear();
     }
 
     @Override
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
@@ -359,29 +359,34 @@ public class AnalyticsClient implements JSONSerializable {
     }
 
     /**
-     * Adds the specified Pinpoint attributes to events to track Campaign/Journey analytics
-     * <p>
-     * You should not use this method as it will be called by the NotificationManager when the app is opened
-     * from a push notification.
-     *
-     * @param eventSourceAttributes the map with pinpointAttributes attributes of the pinpointAttributes received
+     * Replaces the current event source attributes (if any)
+     * with a new set of event source attributes
+     * @param attributes map of attribute values
      */
-    public void setEventSourceAttributes(Map<String, String> eventSourceAttributes) {
-        if (eventSourceAttributes == null) {
-            log.warn("Null pinpointAttributes attributes provided to setPinpointAttributes.");
-            return;
+    public void updateEventSourceGlobally(final Map<String, String> attributes) {
+        //This call removes previous events identifiers from the
+        //globalAttributes collection in AnalyticsEvent.
+        clearEventSourceAttributes();
+        for (final Map.Entry<String, String> entry : attributes.entrySet()) {
+            if (entry.getValue() != null) {
+                addGlobalAttribute(entry.getKey(), entry.getValue());
+            }
         }
-
-        this.eventSourceAttributes = eventSourceAttributes;
+        //We set this so the analytics client knows what to remove from
+        //from the globalAttributes collection when it's time to clear it
+        this.eventSourceAttributes = attributes;
     }
 
     /**
-     * Clears Pinpoint attributes
+     * Removes the attributes currently stored in eventSourceAttributes
+     * from the globalAttributes collection. This is typically called
+     * when we want to switch which event source (campaign, journey, or other future
+     * pinpoint construct) pinpoint events are attributed to.
      * <p>
      * You should not use this method as it will be called by the NotificationManager when the app is opened
      * from a push notification.
      */
-    public void clearEventSourceAttributes() {
+    void clearEventSourceAttributes() {
         for (final String key : eventSourceAttributes.keySet()) {
             this.removeGlobalAttribute(key);
         }
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
@@ -33,7 +33,7 @@ import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.Not
 import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.JOURNEY_ACTIVITY_ID_ATTRIBUTE_KEY;
 import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.JOURNEY_ID_ATTRIBUTE_KEY;
 
-public class EventSourceType {
+class EventSourceType {
     private static final String TAG = EventSourceType.class.getSimpleName();
     private static final String AWS_EVENT_TYPE_OPENED = "opened_notification";
     private static final String AWS_EVENT_TYPE_RECEIVED_FOREGROUND = "received_foreground";
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
@@ -40,6 +40,12 @@ class EventSourceType {
     private static final String AWS_EVENT_TYPE_RECEIVED_BACKGROUND = "received_background";
     private static final String PINPOINT_ATTRIBUTE_KEY = "pinpoint";
     private static final String JOURNEY_ATTRIBUTE_KEY = "journey";
+    private static final String CAMPAIGN_EVENT_SOURCE_NAME = "campaign";
+    private static final String JOURNEY_EVENT_SOURCE_NAME = "journey";
+    private static final String CAMPAIGN_EVENT_SOURCE_PREFIX = "_campaign";
+    private static final String JOURNEY_EVENT_SOURCE_PREFIX = "_journey";
+    static final String UNKNOWN_EVENT_SOURCE_NAME = "unknown";
+
     private final EventSourceAttributeParser attributeParser;
     private String eventSourceName;
     private String eventTypeOpenend;
@@ -49,12 +55,7 @@ class EventSourceType {
     private String eventSourceActivityIdAttributeKey;
     private String eventSourceKeyPrefix;
 
-    private static final String CAMPAIGN_EVENT_SOURCE_NAME = "campaign";
-    private static final String JOURNEY_EVENT_SOURCE_NAME = "journey";
-    static final String UNKNOWN_EVENT_SOURCE_NAME = "unknown";
 
-    private static final String CAMPAIGN_EVENT_SOURCE_PREFIX = "_campaign";
-    private static final String JOURNEY_EVENT_SOURCE_PREFIX = "_journey";
 
     private static EventSourceType CAMPAIGN = new EventSourceType(CAMPAIGN_EVENT_SOURCE_NAME,
             CAMPAIGN_EVENT_SOURCE_PREFIX,
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
@@ -47,13 +47,13 @@ final class EventSourceType {
     private static final String UNKNOWN_EVENT_SOURCE_NAME = "unknown";
 
     private final EventSourceAttributeParser attributeParser;
-    private String eventSourceName;
-    private String eventTypeOpenend;
-    private String eventTypeReceivedForeground;
-    private String eventTypeReceivedBackground;
-    private String eventSourceIdAttributeKey;
-    private String eventSourceActivityIdAttributeKey;
-    private String eventSourceKeyPrefix;
+    private final String eventSourceName;
+    private final String eventTypeOpenend;
+    private final String eventTypeReceivedForeground;
+    private final String eventTypeReceivedBackground;
+    private final String eventSourceIdAttributeKey;
+    private final String eventSourceActivityIdAttributeKey;
+    private final String eventSourceKeyPrefix;
 
     private static EventSourceType CAMPAIGN = new EventSourceType(CAMPAIGN_EVENT_SOURCE_NAME,
             CAMPAIGN_EVENT_SOURCE_PREFIX,
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
@@ -1,0 +1,194 @@
+/**
+ * Copyright 2016-2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package com.amazonaws.mobileconnectors.pinpoint.targeting.notification;
+
+import android.os.Bundle;
+import android.util.Log;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonParser;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSyntaxException;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.CAMPAIGN_ACTIVITY_ID_ATTRIBUTE_KEY;
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.CAMPAIGN_ID_ATTRIBUTE_KEY;
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.CAMPAIGN_PUSH_KEY_PREFIX;
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.JOURNEY_ACTIVITY_ID_ATTRIBUTE_KEY;
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.JOURNEY_ID_ATTRIBUTE_KEY;
+
+public class EventSourceType {
+    private static final String TAG = EventSourceType.class.getSimpleName();
+    private static final String AWS_EVENT_TYPE_OPENED = "opened_notification";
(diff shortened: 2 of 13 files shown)


aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
```diff
@@ -359,33 +359,33 @@ public void removeGlobalMetric(String eventType, String metricName) {
     }
 
     /**
-     * Adds the specified campaign attributes to events to track Campaign Analytic
+     * Adds the specified pinpoint attributes to events to track Campaign/Journey Analytic
```
jamesonwilliams, line 362: Marketing nit: a capital `P` for `Pinpoint`, the proper-noun

(Apply same to all natural language portions of this body of work)

------------
Nit 2, it wasn't your bad, but the last person wrote `Analytic`. That's a word, I guess, but there are multiple analytics at play! Can we make it `analytics`?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
```diff
@@ -359,33 +359,33 @@ public void removeGlobalMetric(String eventType, String metricName) {
...
      * You should not use this method as it will be called by the NotificationManager when the app is opened
      * from a push notification.
      *
-     * @param campaign the map with campaign attributes of the campaign received
+     * @param eventSourceAttributes the map with pinpointAttributes attributes of the pinpointAttributes received
      */
-    public void setCampaignAttributes(Map<String, String> campaign) {
-        if (campaign == null) {
-            log.warn("Null campaign attributes provided to setCampaignAttributes.");
+    public void setEventSourceAttributes(Map<String, String> eventSourceAttributes) {
```
jamesonwilliams, line 369: I'm not sold on the naming for this... "Event source attributes." Possibly, we could just skip the descriptors and get to just `setAttributes()`. or, `setPinpointAttributes()` ... `setGlobalAttributes()` .... are any of these better? Just throwing it out, I'm not sure. Let me know what you think.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Campaign attributes are send from pinpoint flattened
+     * For example:
+     *  "pinpoint.campaign.campaign_id"
+     * Journey attributes come in JSON format
+     * This interface just seeks to abstract some of that away
+     * from the logic that handles the notfication
+     * @return
+     */
+    public interface EventSourceAttributeParser {
```
jamesonwilliams, line 60: Conventionally, nested interfaces/classes come after all other entities, ref, [`InnerTypeLast`](https://checkstyle.sourceforge.io/config_design.html#InnerTypeLast) rule

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+        Map<String, String> getEventSourceAttributes(Bundle bundle);
```
jamesonwilliams, line 61: `get` looks like an accessor. But, your interface is named like it will be ripping stuff from the bundle. A better name for this might be `extractAttributesFromBundle(...)`, or `parseAttributes(...)`, that imply more activity than `get`.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    private static final EventSourceAttributeParser CAMPAIGN_PARSER = new EventSourceAttributeParser() {
```
jamesonwilliams, line 64: It's long enough of a code block that you could consider having `private static final class CampaignAttributeParser implements EventSoruceAttributeParser`, and then just doing `new CampaignAttributeParser()`.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+                    //If it matches the prefix, include it in the attributes
+                    //after stripping out the prefix
+                    campaignAttributes.put(key.replace(campaignKeyPrefix,""), bundle.getString(key));
+                }
+            }
+            return campaignAttributes.size() > 0 ? campaignAttributes : null;
+        }
+    };
+
+    private static final EventSourceAttributeParser JOURNEY_PARSER = new EventSourceAttributeParser() {
```
jamesonwilliams, line 84: Same here, consider splitting to `private static final class JourneyAttributeParser implements EventSourceAttributeParser`.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+     * @return
+     */
+    public interface EventSourceAttributeParser {
+        Map<String, String> getEventSourceAttributes(Bundle bundle);
+    }
+
+    private static final EventSourceAttributeParser CAMPAIGN_PARSER = new EventSourceAttributeParser() {
+        @Override
+        public Map<String, String> getEventSourceAttributes(Bundle bundle) {
+            if(bundle == null) {
```
jamesonwilliams, line 67: See if you can configure your IDE to remind you about these [`WhitespaceAround`](https://checkstyle.sourceforge.io/config_whitespace.html#WhitespaceAround) issues with the keywords, `if, `for`, etc.?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+                return null;
```
jamesonwilliams, line 68: Prefer getting out of `null` as quickly as you can. Prefer return of `@NonNull`, `Collections.emptyMap()`.

Ref a discussion https://stackoverflow.com/q/45009329

(Apply same default return to all location where you return a `null` collection in this PR.)

Or, since your methods are mostly starting with a `new HashMap<>()` and the populating values into it, you could return _that_ object, as your empty collection, instead of the (explicitly stated) `Collections.emptyMap()`.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+    private static final EventSourceAttributeParser JOURNEY_PARSER = new EventSourceAttributeParser() {
+        @Override
+        public Map<String, String> getEventSourceAttributes(Bundle bundle) {
+            String pinpointPayload = bundle.getString(PINPOINT_ATTRIBUTE_KEY);
+            if(pinpointPayload == null) {
+                return null;
+            }
+            Map<String, String> result = null;
+            try {
+                JsonElement pinpointJson = new JsonParser().parse(pinpointPayload);
```
jamesonwilliams, line 93: Limit scope of `try/catch` to just the thing that throws. It _looks_ like that's this `parse(...)` call, but maybe not the stuff that follows (and is still inside the `try`)?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+                Log.v(TAG, "Payload: "+pinpointPayload);
+                return null;
+            }
+
+            return result;
+        }
+    };
+
+    public static EventSourceType CAMPAIGN = new EventSourceType("campaign",
+            "_campaign",
```
jamesonwilliams, line 113: Into a constant? Same for `"_journey"`?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+            CAMPAIGN_PUSH_KEY_PREFIX,
+            CAMPAIGN_PARSER);
+    public static EventSourceType JOURNEY = new EventSourceType("journey",
+            "_journey",
+            JOURNEY_ID_ATTRIBUTE_KEY,
+            JOURNEY_ACTIVITY_ID_ATTRIBUTE_KEY,
+            null,
+            JOURNEY_PARSER);
+
+    public static EventSourceType getEventSourceType(Bundle bundle) {
```
jamesonwilliams, line 125: Can you audit all of these methods -- how many need to be public? Can any be private or default (pakage-local) access?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,194 @@
...
+        if(bundle == null) {
+            return null;
+        }
+        if(bundle.containsKey(CAMPAIGN.getEventSourceKeyPrefix()+CAMPAIGN.getEventSourceIdAttributeKey())) {
```
jamesonwilliams, line 129: `WhitespaceAround` here ... suggest to cache the keycompound to cleanup the `if`,

```
String campaignKey = // form the key compound .../
if (bundle.contains(campaignKey)) {
  // It's a campaign! Wooo!
} else {
  // distinctly not a campaign.
}
```
Same for the next repeated code block int he helper method below.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClient.java
```diff
@@ -284,16 +284,27 @@ public String getChannelType() {
      * local broadcast is sent.
      *
      * @param notificationDetails the notification message received by the device's messaging service
-     * @return {@link NotificationClient.CampaignPushResult}.
+     * @return {@link PinpointPushResult}.
+     *
+     */
+    public PinpointPushResult handlePushNotification(final NotificationDetails notificationDetails) {
+        return notificationClientBase.handlePushNotification(notificationDetails);
```
jamesonwilliams, line 291: Ughh. This isn't your bad, I'm just commenting on the existing architecture here of this legacy code base .... when you have a `Foo`, and a `FooBase`, you'd expect `Foo extends FooBase`, right? But no. Here, `Foo` "has a" `FooBase` ah man.

aws-android-sdk-pinpoint/src/test/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/GCMNotificationClientTest.java
```diff
@@ -152,14 +154,76 @@ private Bundle buildPushBundle() throws JSONException {
         return pushBundle;
     }
 
+    private Bundle buildJourneyPushBundle() throws JSONException {
```
jamesonwilliams, line 157: Can you break this up a bit with empty lines between the construction points?

Also, my faovrite way to bulid readable JSON is:
```
new JSONObject("key", new JSONObject()
    .put("innerKey", new JSONObject()
         .put("innerInnerkey", "innerInnerStringVal")
    )
).toString();
```
The indentation of the key names mirrors an actual JSON document and helps the reader understand the nesting levels with a quick glance.

Same suggestion applies to any place you are constructing JSON in this way.

aws-android-sdk-pinpoint/src/test/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/GCMNotificationClientTest.java
```diff
@@ -152,14 +154,76 @@ private Bundle buildPushBundle() throws JSONException {
...
+        assertThat(receivedEvent.getAttribute("isAppInForeground"), is("false"));
+        assertThat(receivedEvent.getAttribute("journey_id"), is("fake_journey_id"));
+        assertThat(receivedEvent.getAttribute("journey_activity_id"), is("fake_journey_activity_id"));
+        assertThat(receivedEvent.getAttribute("journey_run_id"), is("fake_journey_run_id"));
+        assertThat(receivedEvent.getAttribute("feedback"), is("random_feedback"));
+        assertThat(receivedEvent.getAttribute("endpoint_id"), is("fake_endpoint_id"));
+        assertThat(receivedEvent.getAttribute("user_id"), is("fake_user_id"));
+        // optOut is true because this test can't get the app icon resource id.
+        assertThat(receivedEvent.getAllMetrics().size(), is(0));
+    }
```
jamesonwilliams, line 217: Can you rework this to just:
```
assertEquals(
  expectedHighLevelObject,
  actualHighLevelObject
);
```
which will compare the two via `.equals()` on the objects.

This assumed that `AnalyticsEvent` has a sensible `equals()` implementation. Which, it may not, actually...

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClient.java
```diff
@@ -284,16 +284,27 @@ public String getChannelType() {
...
+    }
+
+    /**
+     * @deprecated see {@link #handlePushNotification(NotificationDetails)}
      */
-    public CampaignPushResult handleCampaignPush(final NotificationDetails notificationDetails) {
-        return notificationClientBase.handleCampaignPush(notificationDetails);
+    @Deprecated
+    public PinpointPushResult handleCampaignPush(final NotificationDetails notificationDetails) {
```
jamesonwilliams, line 298: Since the return type has changed, anyway, the `@Deprecated` isn't helping much. We might as well just delete it.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
rjuliano: The thinking behind this was to keep it consistent with what's being done on the [iOS side](https://github.com/aws-amplify/aws-sdk-ios/pull/2387)

But maybe we don't need to do that??

rjuliano: I'll take a look at the Pinpoint documentation to see if they use some sort of terminology that would be less verbose.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: I like parseAttributes...I'm terrible at naming stuff 😄

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: That's a good point. I was also able to limit the scope by changing that

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: See above

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
jamesonwilliams: You’re not alone!

> There are only two hard things in Computer Science: cache invalidation and naming things.
> 
> -- Phil Karlton

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: I couldn't find a way to get the IDE to flag these. So just setup checkstyle locally to pick these up 😄 I'll have to fiddle with android studio a little more though...should be an easier way to do it through there

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: It looks like the `@NonNull` annotation comes from `com.android.support:support-annotations` dependency. Looking through the SDK, doesn't look like any of the modules take a dependency on that right now. Is that something we can add?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: Done

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: Done

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: I've gone through and there are only a couple things that remain public now.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: Done. I also modified this to return an UNKNOWN event source. Returning journey on else felt a bit misleading.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClient.java
rjuliano: Yeah...that was very confusing. I thought about going through and taking a stab at making this a little cleaner, but the scope started to get a little bit larger. On top of that, they created classes that then inherit from `Foo` for each of the push notification services 🤦‍♂

aws-android-sdk-pinpoint/src/test/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/GCMNotificationClientTest.java
rjuliano: I like that. I went ahead and made the change and also cleaned up the other method that creates the payload for campaign tests

aws-android-sdk-pinpoint/src/test/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/GCMNotificationClientTest.java
rjuliano: Yeah...there's currently no implementation of `equals()` in that class. I came up with one (see below), but I'm not sure it would make sense for this test. In order for it to work for the test, I have to leave out event_id and timestamp which are internally generated; leaving those out of an `equals()` implementation just feels wrong. I'd be creating the override just for the test.

For the purposes of this test, I can probably just change the assertions to:
```
assertTrue(receivedEvent.getEventTimestamp() > 0);
assertEquals("_journey.received_background", receivedEvent.getEventType());
assertEquals(expectedEvent.getAllAttributes(), receivedEvent.getAllAttributes());
```
That should validate everything I care about: the event type, all the parameter keys and values, and that a timestamp was set.

Let me know what you think.

For reference, my implementation of `equals()` for `AnalyticsEvent`
```
@Override
    public boolean equals(Object other) {
        if (other instanceof AnalyticsEvent) {
            AnalyticsEvent analyticsEvent = (AnalyticsEvent) other;
            return this.getEventType().equals(analyticsEvent.getEventType()) &&
                this.getAllMetrics().size() == analyticsEvent.getAllMetrics().size() &&
                this.getAllMetrics().equals(analyticsEvent.getAllMetrics()) &&
                this.getAllAttributes().size() == analyticsEvent.getAllAttributes().size() &&
                this.getAllAttributes().equals(analyticsEvent.getAllAttributes()) &&
                this.getSdkName().equals(analyticsEvent.getSdkName()) &&
                this.getSdkVersion().equals(analyticsEvent.getSdkVersion()) &&
                this.getEventId().equals(analyticsEvent.getEventId());
        } else {
            return false;
        }
    }
```

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
rjuliano: Yup...fixed both of those. I scanned the file for other occurrences as well.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: I actually ended up changing from an interface to a class that had an implementation of the parse method that returns an empty map. There were a lot of places where I was making that assumption that if it wasn't a campaign, I was defaulting to a journey. That felt kind of wrong so I created an `UNKNOWN` event source type.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: I did a little more refactoring to this today and moved it to the bottom

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClient.java
rjuliano: So here's what I did: I added the old enum back and changed `handleCampaignPush` to return it. Still marked it as deprecated.

The new `handlePushNotification` method returns the new enum type with a more generic name. 

I'm going to get an API review in the books to go over this.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClientBase.java
```diff
@@ -826,127 +829,136 @@ private void openURL(final String url, final boolean noSchemeValidation) {
...
         try {
             // Ignore whether the app is in the foreground if the configuration indicates it should post
             // notifications in the foreground.
-            if (
-                !pinpointContext.getPinpointConfiguration().getShouldPostNotificationsInForeground() && isAppInForeground) {
+            if (!pinpointContext.getPinpointConfiguration().getShouldPostNotificationsInForeground() && isAppInForeground) {
                 // Notify the caller that the app was in the foreground.
-                return NotificationClient.CampaignPushResult.APP_IN_FOREGROUND;
+                return NotificationClient.PushResult.APP_IN_FOREGROUND;
             } else {
```
zhinang-amazon, line 919: If the push notification is received in the background, we also want to add these attributes to AnalyticsClient's global attributes. Should we call `addGlobalEventSourceAttributes` here?

rjuliano: I thought we didn't want to count that since the app was already in the foreground and `ShouldPostNotificationsInForeground` is set to `false`. If the `ShouldPostNotificationsInForeground` is set to `true` if would take the else path of this if statement. 

Let me know if this makes sense or if I misunderstood you.

zhinang-amazon: Sorry, I meant the else case, which is when app is in the background.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
rjuliano: I refactored some of the code a bit to make the logic a little clearer (maybe). I moved most of the code that deals with manipulating the attributes into the `AnalyticsClient`. The `NotificationClientBase` methods now just invokes `updateEventSourceGlobally` in `AnalyticsClient`

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClientBase.java
rjuliano: With regards to the concurrency issue when clearing the global attributes, I don't believe that's an issue for us. The global attributes are stored in an in-memory `ConcurrentHashMap`. When events are recorded, we pull those attributes from that data structure and create the event payload.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
```diff
@@ -359,33 +359,38 @@ public void removeGlobalMetric(String eventType, String metricName) {
...
-     * @param campaign the map with campaign attributes of the campaign received
+     * Replaces the current event source attributes (if any)
+     * with a new set of event source attributes
+     * @param attributes map of attribute values
      */
-    public void setCampaignAttributes(Map<String, String> campaign) {
-        if (campaign == null) {
-            log.warn("Null campaign attributes provided to setCampaignAttributes.");
-            return;
+    public void updateEventSourceGlobally(final Map<String, String> attributes) {
```
jamesonwilliams, line 366: Where is this being called from? It's part of the public API, no? If anyone has `pinpointManager.getAnalyticsClient().setCampaignAttributes(...)` in their code base, this would be a breaking change.

Like for the other method, you could _add_ this method, and have it pass to an `@Deprecated public void setCampaignAttributes(....)`.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
```diff
@@ -359,33 +359,38 @@ public void removeGlobalMetric(String eventType, String metricName) {
...
+        for (final Map.Entry<String, String> entry : attributes.entrySet()) {
+            if (entry.getValue() != null) {
+                addGlobalAttribute(entry.getKey(), entry.getValue());
+            }
         }
-
-        campaignAttributes = campaign;
+        //We set this so the analytics client knows what to remove from
+        //from the globalAttributes collection when it's time to clear it
+        this.eventSourceAttributes = attributes;
```
jamesonwilliams, line 377: If `attributes` is _not_ a `ConcurrentHashMap<>()`, you'll lose your assignment type, that you'd chose earlier.

It may be better to make the class field as `final`, and then use `eventSourceAttributes.addAll(attributes);`, instead.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,204 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.CAMPAIGN_ACTIVITY_ID_ATTRIBUTE_KEY;
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.CAMPAIGN_ID_ATTRIBUTE_KEY;
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.CAMPAIGN_PUSH_KEY_PREFIX;
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.JOURNEY_ACTIVITY_ID_ATTRIBUTE_KEY;
+import static com.amazonaws.mobileconnectors.pinpoint.targeting.notification.NotificationClientBase.JOURNEY_ID_ATTRIBUTE_KEY;
+
+class EventSourceType {
```
jamesonwilliams, line 36: `final class`?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,204 @@
...
+    private static final String AWS_EVENT_TYPE_OPENED = "opened_notification";
+    private static final String AWS_EVENT_TYPE_RECEIVED_FOREGROUND = "received_foreground";
+    private static final String AWS_EVENT_TYPE_RECEIVED_BACKGROUND = "received_background";
+    private static final String PINPOINT_ATTRIBUTE_KEY = "pinpoint";
+    private static final String JOURNEY_ATTRIBUTE_KEY = "journey";
+    private static final String CAMPAIGN_EVENT_SOURCE_NAME = "campaign";
+    private static final String JOURNEY_EVENT_SOURCE_NAME = "journey";
+    private static final String CAMPAIGN_EVENT_SOURCE_PREFIX = "_campaign";
+    private static final String JOURNEY_EVENT_SOURCE_PREFIX = "_journey";
+    static final String UNKNOWN_EVENT_SOURCE_NAME = "unknown";
```
jamesonwilliams, line 47: Not `private`?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,204 @@
...
+    private final EventSourceAttributeParser attributeParser;
+    private String eventSourceName;
+    private String eventTypeOpenend;
+    private String eventTypeReceivedForeground;
+    private String eventTypeReceivedBackground;
+    private String eventSourceIdAttributeKey;
+    private String eventSourceActivityIdAttributeKey;
+    private String eventSourceKeyPrefix;
+
+
```
jamesonwilliams, line 58: nit: a couple of extra empty lines

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClient.java
```diff
@@ -277,23 +277,44 @@ public String getChannelType() {
...
     public CampaignPushResult handleCampaignPush(final NotificationDetails notificationDetails) {
-        return notificationClientBase.handleCampaignPush(notificationDetails);
+        return CampaignPushResult.valueOf(notificationClientBase.handleNotificationReceived(notificationDetails).toString());
     }
 
     /**
-     * Result values of handling a pinpoint push message.
+     * @deprecated see {@link PushResult}
      */
     public enum CampaignPushResult {
```
jamesonwilliams, line 305: The Javadoc `@deprecated` is good. But, it we should get a Java `@Deprecated` on the class itself, as well.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
rjuliano: Good call on this one. I'll add this back in and deprecate it.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
rjuliano: Ah...good point. Updating that in the next commit.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: I was referencing this field in one single place to determine if the event type is unknown. I went ahead and added a method called isUnkown that does the comparison and was able to mark this field as private. A little cleaner that way

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
rjuliano: fixed

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/NotificationClient.java
rjuliano: That was a miss on my part. I'm adding that.

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/analytics/AnalyticsClient.java
jamesonwilliams: Arg, actually the existing method had:
```
You should not use this method as it will be called by the NotificationManager when the app is opened
     * from a push notification.
```
Why was this method `public` originally? A customer isn't supposed to actually call it I guess?

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/targeting/notification/EventSourceType.java
```diff
@@ -0,0 +1,206 @@
...
+    private static final String UNKNOWN_EVENT_SOURCE_NAME = "unknown";
+
+    private final EventSourceAttributeParser attributeParser;
+    private String eventSourceName;
+    private String eventTypeOpenend;
+    private String eventTypeReceivedForeground;
+    private String eventTypeReceivedBackground;
+    private String eventSourceIdAttributeKey;
+    private String eventSourceActivityIdAttributeKey;
+    private String eventSourceKeyPrefix;
```
jamesonwilliams, line 56: Can all of these be `final`?

rjuliano: done

PR #1543, opened 2020-04-03
[IoT] Support for custom domain and enhanced custom auth flow

desokroshan:
*Issue #, if available:*

*Description of changes:*
This PR adds changes to support connecting to IoT using custom endpoints. Fully qualified domain name is used as SNI hostname in SSL handshake. This also adds support for authentication by passing credentials in the username and password fields of an MQTT CONNECT message. 

https://docs.aws.amazon.com/iot/latest/developerguide/enhanced-custom-authentication.html
https://docs.aws.amazon.com/iot/latest/developerguide/iot-custom-endpoints-configurable-aws.html


aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -730,6 +774,7 @@ private void initDefaults() {
         connectionStabilityTime = DEFAULT_CONNECTION_STABILITY_TIME_SECONDS;
         unitTestMillisOverride = null;
         needResubscribe = true;
+        this.customWebsocketHeaders = new Properties();
```
jamesonwilliams, line 777: Ordinarily `java.util.Properties` is used to read a `something.properties` file, which contains a list of key-value pairs, separated by `=`. Java `Properties` has been part of the Java ecosystem since 1.0, but is not widely used, today.

For example, a Java properties file looks like:
```
# config.properties
post=8888
host=coolhost.net
```

Android largely does not use these files, or this utility class. I was trying to think _why_ that is, and I'm not entirely sure. But here are two theories.

1. Android included a rich notion of `Resources`, which is more feature-rich and capable than the `Properties`.
2. Android is a younger derivative of Java. Android 1 was built on Java 1.5, and even by Gingerbread (2.3), Android was at 1.6. Point in short: the original practices of Java 1.0 were more distant in memory than the original Android ecosystem.

So anyway - TLDR, I'd use a `Map<String, String>` here, actually. `Properties` models a properties file, and wraps a `Map`. these headers here are literally a dictionary of key/values, without any other interesting behaviors.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -972,19 +1017,58 @@ public void connect(
             final String customAuthorizer,
             final AWSIotMqttClientStatusCallback statusCallback) {
 
-        if (tokenKeyName == null || token == null || tokenSignature == null) {
+        if (tokenKeyName == null || token == null) {
```
jamesonwilliams, line 1020: The stipulation in the documentation is that we will no longer complain about token signature stuff _if_ the user is using a custom endpoint. Do we still have a safeguard here in the case where the user is _not_ using custom auth features?

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -972,19 +1017,58 @@ public void connect(
...
         this.userStatusCallback = statusCallback;
 
+        final MqttConnectOptions options = new MqttConnectOptions();
+        // Set custom websocket header
+        customWebsocketHeaders.setProperty("X-Amz-CustomAuthorizer-Name", customAuthorizer);
+        customWebsocketHeaders.setProperty(tokenKeyName, token);
+        if (tokenSignature != null) {
+            customWebsocketHeaders.setProperty("X-Amz-CustomAuthorizer-Signature", tokenSignature);
+        }
+        customAuthConnect(options);
```
jamesonwilliams, line 1033: This is a little confusing. You make an `options`, do some other stuff that doesn't touch `options`, and then call `customAuthConnect(options)`. 

Does that method use the headers you've just setup? To make that fact more clear, I wonder if you can just pass the map into the `customAuthConnect(...)` method, as a param?

Or better, should the headers be affixed into the `options`, somehow? Like maybe you do `options.setCustomWebsockHeaders(Map<String, String> headers)`, and then proceed to pass the `options` as-is, to the `customAuthConnect(` method?

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -972,19 +1017,58 @@ public void connect(
...
+    }
+
+    /**
+     * Initializes the MQTT session and connects to the specified MQTT server
+     * using username and password as credentials.
+     *
+     * @param username Username to authenticate client to the broker.
+     * @param password Password to authenticate client to the broker.
+     * @param statusCallback When a new MQTT session status is received, a function of callback will
```
jamesonwilliams, line 1042: wording, `a function of callback will be called` --> `this callback function will be invoked`

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -972,19 +1017,58 @@ public void connect(
...
+            final String username,
+            final String password,
+            final AWSIotMqttClientStatusCallback statusCallback) {
+
+        if (username == null || password == null) {
+            throw new IllegalArgumentException("username and/or password cannot be null.");
+        }
+
+        this.username = username;
+        this.password = password;
```
jamesonwilliams, line 1055: If you pass these into the methods below, do you need to cache these into the class? Or can they just be method-scoped?

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -972,19 +1017,58 @@ public void connect(
...
+        this.userStatusCallback = statusCallback;
```
jamesonwilliams, line 1056: Same here, do you need to cache this on the class? Or can you just pass it as an argument to `customerAuthConnect(options, statusCallback)`? 

Currently, have you have `null` guards on the `username` and `password`, but not on this `statusCallback`. Is this `connect(...)` method expected to work, if the user passes `null` for the callback? If not, you could add a guard for it too.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -1176,6 +1253,10 @@ void reconnectToSession() {
                 case CUSTOM_AUTH:
                     options.setCustomWebSocketHeaders(customWebsocketHeaders);
                     break;
+                case USERNAME_PASSWORD:
+                    options.setUserName(username);
```
jamesonwilliams, line 1257: Is this duplicative of the work done around line 1059?

aws-android-sdk-iot/src/test/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManagerTest.java
```diff
@@ -273,6 +273,31 @@ public void testConnectWebSocket() throws Exception {
...
+        mockClient.mockConnectSuccess();
+
+        // Then: connect is invoked on the paho client and client status transitions through connecting
+        // to connected.
+        assertEquals(1, mockClient.connectCalls);
+        assertTrue(mockClient.mostRecentOptions.isCleanSession());
+        assertEquals(300, mockClient.mostRecentOptions.getKeepAliveInterval());
+        assertEquals(2, csb.statuses.size());
+        assertEquals(AWSIotMqttClientStatusCallback.AWSIotMqttClientStatus.Connecting, csb.statuses.get(0));
+        assertEquals(AWSIotMqttClientStatusCallback.AWSIotMqttClientStatus.Connected, csb.statuses.get(1));
```
jamesonwilliams, line 297: I'd include an import for this static inner entity, `AWSIotMqttClientStatus`. Next, you can simplify these three assertions into one:
```
assertEquals(
    Arrays.asList(AWSIotMqttClientStatus.Connecting, AWSIotMqttClientStatus.Connected),
    csb.statuses
);
```

aws-android-sdk-iot/src/test/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManagerTest.java
```diff
@@ -273,6 +273,31 @@ public void testConnectWebSocket() throws Exception {
...
+    public void testWebsocketConnectionWithUsernamePassword() throws Exception {
+        // Given: the MqttClient and an instance of AWSIotMqttManager.
+        MockMqttClient mockClient = new MockMqttClient();
+        AWSIotMqttManager awsIotMqttManager = new AWSIotMqttManager("test-client",
+                Region.getRegion(Regions.US_EAST_1), TEST_ENDPOINT_PREFIX);
+        awsIotMqttManager.setMqttClient(mockClient);
+        TestClientStatusCallback csb = new TestClientStatusCallback();
+
+        // When: the SDK tries to connect to IoT.
+        awsIotMqttManager.connect("user", "password", csb);
```
jamesonwilliams, line 286: Instead of waiting an arbitrary amount of time, wait for the even of interest -- the callback occurring. Perhaps, something like:
```
CountDownLatch latch = new CountDownLatch(1);
awsIotMqttManager.connect("user", "password", new TestClientStatusCallback() {
    @Override
    public void onSomeStatusEventCallback(StatusInputType type) {
        super.onSomeStatusEventCallback(type); // Same behavior so far...
        latch.countDown();
    }
});
latch.await(500, TimeUnit.MILLISECONDS);
```
^ In this case, the arbitrary value is the _upper bound_ for the result. But, most of the time, the result will be rendered _before_ 500ms has elapsed.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotWebSocketUrlSigner.java
```diff
@@ -57,18 +57,43 @@ public AWSIotWebSocketUrlSigner(String serviceName) {
...
+     *
+     * @param endpoint service endpoint with or without customer specific URL prefix.
+     * @param awsCredentials credential set to be used in SigV4 signature algorithm.
+     * @param currentTimeInMillis time value to be used in SigV4 calculations.  In milliseconds.
+     * @return a URL with SigV4 signature formatted to be used with AWS IoT.
+     */
+    String getSignedUrl(final String endpoint,
+                        final AWSCredentials awsCredentials,
+                        final long currentTimeInMillis) {
+        return getSignedUrl(endpoint, awsCredentials, currentTimeInMillis, null);
```
jamesonwilliams, line 69: Instead of passing `null`, and then checking if its null, just ensure that it's  never null, to begin.

To achieve this, you can have this line look like:
```
Region endpointRegion = AwsIotEndpointUtility.getRegionFromIotEndpoint(endpoint);
return getSignedUrl(endpoint, awsCredentials, currentTimeInMillis, endpointRegion);
```

Then, in your new overloaded version of this method, just always use the `region`, without any conditional logic.

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -767,6 +770,14 @@ public void onStatusChanged(AWSIotMqttClientStatus status, Throwable throwable)
                     statuses.add(status);
                 }
             });
+        } else if (AWSIotMqttManager.AuthenticationMode.USERNAME_PASSWORD.equals(authMode)) {
```
jamesonwilliams, line 773: Now that this control structure has moved from a ~simple `if/if-else`, you should probably move to a `switch` on `authMode`.

aws-android-sdk-iot/src/androidTest/java/com/amazonaws/mobileconnectors/iot/MqttManagerIntegrationTest.java
```diff
@@ -849,6 +866,14 @@ public void onStatusChanged(AWSIotMqttClientStatus status, Throwable throwable)
                             statuses.add(status);
                         }
                     });
+        } else if (AWSIotMqttManager.AuthenticationMode.USERNAME_PASSWORD.equals(authMode)) {
```
jamesonwilliams, line 869: Ditto here

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
desokroshan: I agree but Paho uses `java.util.Properties` for all of their key/value storage needs.
https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.client.mqttv3/src/main/java/org/eclipse/paho/client/mqttv3/MqttConnectOptions.java#L76

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
desokroshan: Good point. Token-signature is optional for all types of endpoints not just custom ones, if token-signing is disable in the domain configuration. Old, existing IoT endpoint however do not support that as yet. Will make a note about it in the comment as discussed.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
desokroshan: These need to be set again for reconnection in case initial attempt at connection fails or an existing connection drops due to network issue. At this point it feels like I should just cache options object at class level and method-scope different fields of options object that are used by different flavors of connect method. I can do a separate PR to refactor this manager class since I think that might be little voluminous.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
desokroshan: `statuscallback` is allowed to be null and we notify the user only if a non-null status callbacks is provided. This callback is being invoked for status change notification from connects, disconnect and reconnect and is therefore class-scoped.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
desokroshan: This is setting options for reconnection as opposed to 1059 where it was for first connection attempt.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
jamesonwilliams: Got it, makes sense!

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -1772,6 +1880,66 @@ static boolean isTopicMatch(String topicFilter, String topic) {
...
+            this.value = value;
+        }
+
+        /**
+         * Static factory method to instantiate {@link Endpoint} from {@link String} value.
+         *
+         * @param value a {@link String} value for mqtt broker endpoint.
+         * @return an instance of {@link Endpoint}.
+         */
+        public static Endpoint froString(final String value) {
```
jamesonwilliams, line 1929: Typo, `froString` --> `fromString`

desokroshan: Ah, thanks for the catch. Fixed!

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -972,19 +1038,65 @@ public void connect(
...
+        }
 
+        this.username = username;
+        this.password = password;
+        userStatusCallback = statusCallback;
+        authMode = AuthenticationMode.USERNAME_PASSWORD;
+        customAuthConnect(getOptions());
+    }
+
+    private MqttConnectOptions getOptions() {
```
jamesonwilliams, line 1082: `getFoo()` sounds like an accessor to a cached value, this  should be named like a factory method, i.e., `createOptions()`, or `buildOptions()`, or `generateOptions()` whatever.

aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AWSIotMqttManager.java
```diff
@@ -707,11 +718,64 @@ public AWSIotMqttManager(String mqttClientId, Region region, String accountEndpo
...
+        if (TextUtils.isEmpty(mqttClientId)) {
+            throw new IllegalArgumentException("mqttClientId is null or empty");
+        }
+
+        if (region == null) {
+            throw new IllegalArgumentException("region is null");
+        }
+
+        if (endpoint == null) {
+            throw new IllegalArgumentException("endpoint is null or an invalid domain.");
```
jamesonwilliams, line 749: Might as well use `TextUtils.isEmpty(...)` here, too, I guess. Either way would update the error message, the "or an invalid domain." part is not actually checked, right?

Issue #1493, opened 2020-02-28
Pass clientMetaData to Cognito Triggers


royjit:
Add clientMetaData into the Auth apis.

Similar ticket in iOS - https://github.com/aws-amplify/aws-sdk-ios/issues/2299

Skyfallz:
Hi @royjit,

as the iOS version seems [merged](https://github.com/aws-amplify/aws-sdk-ios/pull/2328), is it possible to get an update about the Android version ? Not passing metadata is blocking for implementing properly Custom Auth Workflows.

Thanks,

Skyfallz:
Hi,

any update @royjit ?

The iOS version was released, and we really need this on the Android version as well.

Thanks,

tys-bradford-zava:
+1 this is also something our team would really need soon on the Android side to match what we have on our iOS app.

mobileappconsultant:
This feature request was put in almost a week ago.. how far are we from it being implemented?

pedroaiwin:
Hi @royjit,

As in the previous comments, our team need to add this feature to the Android project, since it is already implemented in iOS.

We have created a new issue [#2547](https://github.com/aws-amplify/aws-sdk-ios/issues/2547) on the iOS repository in order to add the clientMetaData parameter to the [resendSignUpCode](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_ResendConfirmationCode.html) call, it is possible to take it into account for the android version?

Thanks

hama0511:
We developed the web using amplify.

However, because the clientMetadata cannot be added in the signUp call in Android, we can no longer proceed with Android app development.

I'm so angry. All code we have developed regarding user authentication should be deleted.

I can't trust aws anymore.

jpignata:
Heya @hama0511 -- this functionality was added to our signup APIs in https://github.com/aws-amplify/aws-sdk-android/pull/1547 butthis issue wasn't properly closed out. Are you encountering an issue trying to pass metadata to this method?

jamesonwilliams:
Hi @hama0511! Sorry you're hitting a snag. As @jpignata mentions, I think we do have support for this. We'd like to work with you to resolve your problems, if we can. Feel free to join us on our Discord server to discuss: https://discord.gg/amplify. We're in #android-help.

hama0511:
Sorry! I found it AWSMobileClient.
But not Amplify..
should I use AWSMobileClient instead amplify?

[...]

PR #1547, opened 2020-04-06
[AWSMobileClient] Update mobile client apis to accept clientMetaData field

desokroshan:
*Issue #, if available:*
https://github.com/aws-amplify/aws-sdk-android/issues/1493

*Description of changes:*
Updated the following Cognito APIs to accept client meta data
[signUp](https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java#L1849)
[confirmSignUp](https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java#L1941)
[forgotPassword](https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java#L2048)
[confirmForgotPassword](https://github.com/aws-amplify/aws-sdk-android/blob/master/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java#L2107)


mobileappconsultant:
Guys is it ok to pull this?

desokroshan:
> Guys is it ok to pull this?

@mobileappconsultant We are working on it actively and will try to get this in as soon as possbile.

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUserPool.java
```diff
@@ -432,12 +472,13 @@ public void signUp(final String userId, final String password,
      * @param userAttributes    REQUIRED: User attributes.
      * @param validationData    REQUIRED: Validation key value pairs, these will be passed to pre
      *                          and post registration lambda functions.
-     *
+     * @param clientMetadata    REQUIRED: Client metadata for lambda function for user registration
```
TrekSoft, line 475: This should not be marked as REQUIRED since it isn't

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -2044,29 +2134,72 @@ public void onFailure(Exception exception) {
...
+    /**
+     * Used to reset password if user forgot the old password.
+     *
+     * @param username username of the user trying to reset password.
+     * @param clientMetadata meta data to be passed to the lambdas invoked by confirm sign up operation.
+     * @param callback callback will be invoked to notify the success or failure of the
+     *                 forgot password operation
+     */
+    @AnyThread
+    public void forgotPassword(final String username,
```
TrekSoft, line 2146: Do we need clientMetadata for forgotPassword?

TrekSoft:
Looks good to me besides a couple comments.

@jamesonwilliams in this case I disagree about replacing null with an enum. This client metadata map is an optional parameter which can be added as extra data to the request or not so I think it's legitimate what he's doing here offering overloaded versions of the method - one which takes it as a parameter and one which doesn't and having the one which doesn't call the one which does with a null value.

Also I didn't see what part you were talking about which uses true/false as a toggle. Which section were you referring to?

jamesonwilliams:
> Looks good to me besides a couple comments.
> 
> @jamesonwilliams in this case I disagree about replacing null with an enum. This client metadata map is an optional parameter which can be added as extra data to the request or not so I think it's legitimate what he's doing here offering overloaded versions of the method - one which takes it as a parameter and one which doesn't and having the one which doesn't call the one which does with a null value.
> 
> From a customer standpoint, if they want to pass it in they call the version which takes it as a parameter and if they don't want to pass it, they call the overloaded version which doesn't take it as a parameter.
> 
> Also I didn't see what part you were talking about which uses true/false as a toggle. Which section were you referring to?

Hey David, for that map, I wouldn't suggest anything with an `enum`. But, I would suggest using `Collections.emptyMap()` in place of `null`.

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
desokroshan: ForgotPassword operation can potentially invoke pre-signup, user migration and custom message lambda triggers. Based on that custom logic in those lambdas, user might need to pass inputs to those lambda using client meta data. This was also the part of original request https://github.com/aws-amplify/aws-sdk-ios/issues/2299

desokroshan:
> > Looks good to me besides a couple comments.
> > @jamesonwilliams in this case I disagree about replacing null with an enum. This client metadata map is an optional parameter which can be added as extra data to the request or not so I think it's legitimate what he's doing here offering overloaded versions of the method - one which takes it as a parameter and one which doesn't and having the one which doesn't call the one which does with a null value.
> > From a customer standpoint, if they want to pass it in they call the version which takes it as a parameter and if they don't want to pass it, they call the overloaded version which doesn't take it as a parameter.
> > Also I didn't see what part you were talking about which uses true/false as a toggle. Which section were you referring to?
> 
> Hey @TrekSoft, for that map, I wouldn't suggest anything with an `enum`. But, I would suggest using `Collections.emptyMap()` in place of `null`.
> 
> The `boolean`s are found in the `confirmSignUp*(...)` family of methods. One is named like `forceOverrideSomeProperty`, or something like that. My argument is that `PropertyBehavior.OVERRIDE` would be more readable and discoverable in code, than `true`.

@jamesonwilliams I have updated the PR to use empty maps in place of nulls.
The Boolean `forcedAliasCreation` to ConfirmSignUp class is part of service API and we do not toggle operations based on that in the SDK code.

desokroshan:
> Not a huge deal, but doesn't `Collections.emptyMap()` undergo type inference to become a `Map<String, String>`, if you pass it into a typed argument, like you are?

It doesn't actually, probably because we are using JDK 1.6 which does not have required type inference

jamesonwilliams:
> Pending the tests pass after the branch is restored I'm good with this - I still don't think passing an empty map is the right way to go on this since up until now we've been passing nothing for this field so passing an empty map is actually changing the behavior of what we're passing to the service but if the tests are fine with it and @jamesonwilliams feels strongly about it I'm fine with that.

Hm this is a good observation. So long as the server treats them the same way, I think we're good. 

I agree that there is an ideological difference between having a plate with 10 green beans, a plate that's empty cause you ate the beans, and then having no plate at all.

Theoretically, we could add a:
```
if (!collection.isEmpty()) {
    requestBuilder.useIt(collection);
} else {
    // don't populate the field at all; leave it out of the request JSON.
    // requestBuilder.useIt(collection);
}
```

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
mobileappconsultant: hey guys.. what's the status of this PR please?

tys-bradford-zava:
Hi all, any movement on this one? This PR is something our team is looking to use as soon as it's available. It's currently a blocker, since we already implemented the iOS equivalent which is now in production.

Thanks

TrekSoft:
Hey, went ahead and merged it into develop here so it should go out with the next release - probably within a couple weeks.

pedroaiwin:
Hi @TrekSoft,

Our team is waiting since the day it was merged on May 14. It is important for us since we want to put the Android app at the same level as the iOS app, which was already released in production with these features.

When will it be possible to have this functionality merged in production?

Thank you

jpignata:
@pedroaiwin Thanks for being patient. We're working on a release now and expect it in the next couple of business days.

PR #1567, opened 2020-04-14
[aws-android-sdk-testutils] Remove testconfiguration.json template

jamesonwilliams:
This file is listed in `.gitignore`, but a templated version exists in the
committed source code. In order for changes to this file to _actually_
be ignored from versioning, this file must not be tracked by git.

This means that `R.raw.testconfiguration` will no longer be available at
compile time. To address this, the AWSTestBase is modified to lookup the
configuration file by its name, at runtime.


raphkim:
I believe `testconfiguration.json` was initially added to allow people to run integration tests with their own resources. Now that we are completely removing the template, the customers will go from having about 10% of the idea to 0%. Is there any alternative way we can achieve that use-case?

jamesonwilliams:
> I believe `testconfiguration.json` was initially added to allow people to run integration tests with their own resources. Now that we are completely removing the template, the customers will go from having about 10% of the idea to 0%. Is there any alternative way we can achieve that use-case?

I agree with the categorization that this file gives (generously!) 10% of the idea. But it is woefully insufficient to guide a contributor to build out the needed backend resources for the tests. If we are honest with ourselves, this file is not helping.

So, we have that problem, either way. Meanwhile, this template file is a ticking time bomb. The presence of this file overrides the rules in `.gitignore`, meaning that this file _is_  versioned. Any changes you make to - such as adding real credentials - can be added to a commit, inadvertently. 

The solution we need to provide, I believe, is a cloud-formation template or something like that, so that anyone (you, me, a CI/CD docker container, etc.) can do a "1-click" and get the test environment without additional domain knowledge, and without special access to privileged materials/credentials.

PR #1570, opened 2020-04-14
Publish SDK artifacts as aar not jar

jamesonwilliams:
The Android archive (aar) format has been around for a long time. The
Android SDK originated from a fork of the AWS Java SDK, which was
publishing Java archives (jar). While it possible to consume jar files
on Android, it is not idiomatic to the platform. Instead, we would like
to publish Android libraries.

This change may require consumers to update their `build.gradle`, in such
cases as they have '@jar' notation in their dependency specifications.


raphkim:
This is a chore that we have always been hoping to get to one day. Thanks for doing this!

jamesonwilliams:
This work will not be pursued further at this time.

jamesonwilliams:
Given https://github.com/aws-amplify/aws-sdk-android/issues/1914 and other issues, I am re-opening and reprioritizing this work.

PR #1598, opened 2020-04-20
Fix incorrect null-check

tomykaira:
*Issue #, if available:*

N/A

*Description of changes:*

This patch fixes the null check where the target object is different from the object that is treated in the following section.


jamesonwilliams:
Hi @tomykaira Is this PR still active? If so, can you update as necessary to resolve conflicts?

jamesonwilliams:
Closing as this has gone stale

PR #1614, opened 2020-04-22
firewallmanagementservice Update models to latest

awsmobilesdk:
firewallmanagementservice Update models to latest


desokroshan:
Not supported yet.

PR #1616, opened 2020-04-22
amazonelasticsearchservice Update models to latest

awsmobilesdk:
amazonelasticsearchservice Update models to latest


desokroshan:
Not supported yet.

PR #1619, opened 2020-04-22
[aws-android-sdk-testutils] Discontinue Use of Long-Lived Credentials

jamesonwilliams:
The tests will no longer use long-lived credentials.

Instead, the `testconfiguration.json` will make reference to a
three-tuple of Access ID, Secret Key ID, Session Token. These are
provided via `BasicSessionCredentials`, from an
`AWSCredentialsProvider`.

The `"Credentials"` block of the `testconfiguration.json` will now require
this form:
```
{
    ...
    "Credentials": {
        "accessKey": "REDACTED_ACCESS_KEY",
        "secretKey": "REDACTED_SECRET_KEY",
        "sessionToken": "REDACTED_SESSION_TOKEN"
    },
    ...
}
```

The details of generating a suitably formatted `testconfiguration.json`
are handled externally to this code base.

Refer: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html

### DIFF
--- a/aws-android-sdk-testutils/src/main/java/com/amazonaws/testutils/AWSTestBase.java
+++ b/aws-android-sdk-testutils/src/main/java/com/amazonaws/testutils/AWSTestBase.java
@@ -25,7 +25,7 @@ import com.amazonaws.AmazonServiceException;
 import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.auth.AWSCredentialsProviderChain;
-import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.auth.BasicSessionCredentials;
 import com.amazonaws.util.StringUtils;
 
 import org.json.JSONException;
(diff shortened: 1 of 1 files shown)


aws-android-sdk-testutils/src/main/java/com/amazonaws/testutils/AWSTestBase.java
```diff
@@ -132,103 +121,81 @@ public String getAccountId() {
...
+
+    public static String getSessionToken() {
+        return getJSONConfiguration().getSessionToken();
+    }
+
     public static String getAccountId() {
         return getJSONConfiguration().getAccountId();
     }
+
     public static JSONObject  getPackageConfigure(String packageName) {
```
phani-srikar, line 191: What is the structure of testconfiguration.json file? Is it like this?
`{ accessKey: "", 
    secretKey: "",
    sessionToken: "",
    Packages: {
         <package ID> : {
              <resource ID> : <resource value>
         }
    }
}`
To keep things consistent across platforms, it's good to have a common "package ID" and a strategy to name "resource ID"

jamesonwilliams: Hey @phani-srikar -- the existing structure is like:
```
{
  "Credentials": {
    "accessKey": "asdfasdf",
    "secretKey": "asdfasdf",
    "sessionToken": "asdfasdf",
    "accountId": "asdfasdf"
  },
  "Packages": {
    "iot": {
       ...
    }
  }
}
```
I just added the `"sessionToken"` for this PR. 

I believe we do have the package ID and resource ID keys already.

phani-srikar: do we need "region" key too in the "credentials" section which would act as default region for all tests unless an override is specified for a package.

PR #1622, opened 2020-04-23
awsresourceaccessmanager Update models to latest

awsmobilesdk:
awsresourceaccessmanager Update models to latest


desokroshan:
Not supported yet.

PR #1625, opened 2020-04-23
awselementalmediapackagevod Update models to latest

awsmobilesdk:
awselementalmediapackagevod Update models to latest


desokroshan:
Not supported yet.

PR #1631, opened 2020-04-24
amazonelasticinference Update models to latest

awsmobilesdk:
amazonelasticinference Update models to latest


desokroshan:
Not supported yet.

PR #1636, opened 2020-04-26
Fixes for S3 tests

jpignata:
*Issue #, if available:* n/a

*Description of changes:*

- PauseTransferIntegrationTest.testMultiPartUploadPause
  - This pause test was intermittently failing both in Android Studio and
        when run through gradlew; it appears that 100 milliseconds wasn't
        enough time for the transfer to transition to CANCELED, which
        resulted in intermittent exceptions. This change raises the wait for
        the transition to CANCELED to a second through the completed latch.
    
 - AclIntegrationTest.testCannedAcls
    - This test had an assertion against a hardcoded value looking at the S3
        display name. It expected the value in the old account this thing ran
        in. There's actually no great way to get this kind of information from
        CloudFormation, and I don't really understand what this is testing (it's
        default S3 behavior to leave existing ACLs in place), so I'm just
        removing the assertion.
    
  - S3IntegrationTest.testGetS3AccountOwner
    - Similarly, this test was looking at the same data from a different
        method and had another hard-coded constant. This isn't a terribly
        well-trod path of AWS these days, there's no easy way to snake this
        through CloudFormation, and I'm not sure it changes our test coverage
        by not having a test for this explicitly.
    
  - BucketReplicationIntegrationTest
    - This test covered the setup of S3 bucket replication, an uncommon
        use-case for mobile developers, and used a hard-coded role to do it.
        Instead of creating and snaking the role through, this change @Ignores
        it for now.

### DIFF
--- a/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/mobileconnectors/s3/transferutility/PauseTransferIntegrationTest.java
+++ b/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/mobileconnectors/s3/transferutility/PauseTransferIntegrationTest.java
@@ -113,7 +113,7 @@ public final class PauseTransferIntegrationTest extends S3IntegrationTestBase {
 
         // cancel early to avoid having to wait for completion
         util.cancel(observer.getId());
-        completed.await(100, TimeUnit.MILLISECONDS);
+        completed.await(1000, TimeUnit.MILLISECONDS);
     }
 
     private final class TestListener implements TransferListener {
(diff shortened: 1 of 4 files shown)


aws-android-sdk-apigateway-test/.idea/workspace.xml
```diff
@@ -0,0 +1,46 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jamesonwilliams: For comparison, we ignore this file in [`amplify-android`'s `.gitignore`](https://github.com/aws-amplify/amplify-android/blob/master/.gitignore#L41). Maybe we should in `aws-sdk-android`, too?

PR #1660, opened 2020-04-28
amazonroute53 Update models to latest

awsmobilesdk:
amazonroute53 Update models to latest


desokroshan:
amazonroute53 not supported.

PR #1674, opened 2020-04-29
Consume snakeCase keys out of testconfiguration.json.

jamesonwilliams:
The `testconfiguration.json` format has changed to use `"credentials"` and
`"packages"` at the top level of the JSON object.

The client needs to expect and read these values in snakeCase, not
PascalCase, as it currently does.


aws-android-sdk-testutils/src/main/java/com/amazonaws/testutils/AWSTestBase.java
```diff
@@ -80,19 +80,19 @@ public JSONObject getPackageConfigure(String packageName) {
         }
```
jpignata, line 80: Do we need to change line 71 too?

jamesonwilliams: Ugh, yes! Good catch.

PR #1700, opened 2020-05-05
Add script to generate config file

jpignata:
*Issue #, if available:* m/a

*Description of changes:* Add script to generate config file

We are generated a test configuration prior to running integration tests
using temporary credentials stored in the environment and configuration
properties stored in AWS Systems Manager Parameter Store. This commit:

- Adds a script to generate the file
- Creates a gradle task to run that script
- Adds a step to CircleCI in order to run the script

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,6 @@
 *.class
 
 # Generated files
-bin/
 gen/
 out/
 
--- a/bin/build_test_config.sh
+++ b/bin/build_test_config.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+DEST_DIR=./aws-android-sdk-testutils/src/main/res/raw
+FILENAME=testconfiguration.json
+REPO=https://github.com/aws-amplify/amplify-ci-support
+SCRIPT=support/src/integ_test_resources/common/device_config_builder.py
+TMP_DIR=$HOME/.aws-amplify
+
+# Make directories if they don't exist
+mkdir -p "$TMP_DIR" "$DEST_DIR"
+
+# Remove previous repo
+rm -rf "${TMP_DIR}/support"
+
+# Clone the support repo into the temporary directory
+(cd "$TMP_DIR" && git clone "$REPO" support)
+
+# Generate file
+(cd "$TMP_DIR" && python "$SCRIPT" android > "$FILENAME")
+
+# Create symbolic link
+ln -f -s "${TMP_DIR}/${FILENAME}" "${DEST_DIR}/${FILENAME}"
--- a/build.gradle
+++ b/build.gradle
@@ -21,3 +21,6 @@ task clean(type: Delete) {
     delete rootProject.buildDir
 }
 
+task buildTestConfig(type:Exec) {
+  commandLine 'bin/build_test_config.sh'
+}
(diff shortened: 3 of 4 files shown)


build.gradle
```diff
@@ -21,3 +21,6 @@ task clean(type: Delete) {
     delete rootProject.buildDir
 }
 
+task buildTestConfig(type:Exec) {
+  commandLine 'bin/build_test_config.sh'
+}
```
jamesonwilliams, line 26: Potentially -- towards benefit of the developer use case -- you could declare this as a dependency for the instrumented tests task. Maybe something like this would work?
```
connectedAndroidTest.dependsOn buildTestConfig
```

bin/build_test_config.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -e
+
+DEST_DIR=./aws-android-sdk-testutils/src/main/res/raw
```
jamesonwilliams, line 5: Currently this script is written in such a way that it needs to be executed from the directory that owns it (use of `./`, etc.)

You could find the `$SCRIPT_PATH` with something like [this](https://stackoverflow.com/a/3572105/695787) and then `dirname $(realpath "$0")`, and base your relative paths against that.

build.gradle
```diff
@@ -21,3 +21,6 @@ task clean(type: Delete) {
     delete rootProject.buildDir
 }
 
+task buildTestConfig(type:Exec) {
+  commandLine 'bin/build_test_config.sh'
```
palpatim, line 25: The script I'm working on will generate the test config file to `$HOME/aws-amplify/aws-sdk-android.testconfiguration.json`. Consider having two steps: one to generate the config file and one to link it, so it'll be easier to swap out the generation piece.

.gitignore
```diff
@@ -11,7 +11,6 @@
 *.class
 
 # Generated files
-bin/
```
jamesonwilliams, line 14: We have this [`CircleCiScripts`](https://github.com/aws-amplify/aws-sdk-android/tree/master/CircleciScripts) dir. Should we re-use that? Possibly and additionally, `git mv CircleCiScripts scripts`; something like that?

palpatim, line 14: Consider a `build-support` or similar directory, rather than removing `bin` from .gitignore. That way we don't accidentally start committing artifacts at some point in the future when we decide it's a good idea for Amplify to generate binaries :)

jpignata: I like the second suggestion!

build.gradle
jpignata: (I had no idea what I was doing for this part)

jpignata:
Letting this cook while Tim works on his script, then I'll try again.

build.gradle
jpignata: @palpatim I'd consider using a `.` directory for these files to reduce the obvious litter. Sounds good otherwise!

build.gradle
jamesonwilliams: I always get sandbagged on this, whenever I have to add task dependencies. Seems like it's never as straightforward as it ought to be.

jpignata:
Closing this in favor of the work happening in https://github.com/aws-amplify/aws-sdk-ios/pull/2545

PR #1908, opened 2020-06-19
Remove allowBackup="true" from AndroidManifest.xml

pyricau:
The AndroidManifest.xml in aws-android-sdk-cognitoauth has `allowBackup=true`. That decision should not belong to a dependency, and having this entry creates friction for consumers of the SDK as it can generate build errors:

```
Manifest merger failed : Attribute application@allowBackup value=(false) from [:common:mymodule] AndroidManifest.xml:37:9-36
        is also present at [com.amazonaws:aws-android-sdk-cognitoauth:2.16.12] AndroidManifest.xml:4:19-45 value=(true).
        Suggestion: add 'tools:replace="android:allowBackup"' to <application> element at AndroidManifest.xml:4:3-5:41 to override.
```

While consumers can use `tools:replace`, it's best to just not have this and avoid the issue in the first place.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,7 +122,7 @@ commands:
           command: |
             if [ -z "$target_branch" ]
             then 
-              target_branch="master"
+              target_branch="main"
             fi
 
             git config --local user.name "${GITHUB_BUMPVERSION_USER}"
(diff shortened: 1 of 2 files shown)


pyricau:
cc @TrekSoft  (sorry if opening a PR already notifies you, I'm not familiar with the customs of this repo :) )

TrekSoft:
Thanks for the PR! We're working on getting deployment issues resolved here and once we do can get this merged in.

PR #1959, opened 2020-06-26
Fixes AWSMobileClientIntegrationTests

TrekSoft:
- Removes some manual tests which were being ignored and causing errors because they required a minSDK version of 23. 
- Fixes a wrong email address which wasn't being tested before because it was never hitting that block of code
- Fixes a wrong auth challenge compared to the comment documenting what Lambda function to create
- Ignores one developer federated identity test as it's an edge case and would be quite difficult to replicate.


aws-android-sdk-mobile-client/src/androidTest/java/com/amazonaws/mobile/client/AWSMobileClientCustomAuthTest.java
```diff
@@ -216,8 +216,8 @@ public void onError(Exception e) {
     private void confirmSignIn() {
         final Map<String, String> res = new HashMap<String, String>();
         final Map<String, String> clientMetaData = new HashMap<String, String>();
-        clientMetaData.put("challengeResponseAnswer", "1134");
-        res.put(CognitoServiceConstants.CHLG_RESP_ANSWER, "1134");
+        clientMetaData.put("challengeResponseAnswer", "1133");
+        res.put(CognitoServiceConstants.CHLG_RESP_ANSWER, "1133");
```
raphkim, line 220: is this just some arbitrary value? Why not use a random number generator for this? 
Or use our `RandomString` utility from `testutils` package (if any string works)?

aws-android-sdk-mobile-client/src/androidTest/java/com/amazonaws/mobile/client/AWSMobileClientPersistenceWithRestartabilityTest.java
```diff
@@ -492,12 +492,12 @@ private void mockHostedUISignIn() throws JSONException {
                 "CognitoIdentityProviderCache",
                 true);
 
-        final Set<String> scopes = new HashSet<String>(Arrays.asList("profile", "openid", "email"));
+        final Set<String> scopes = new HashSet<String>(Arrays.asList("aws.cognito.signin.user.admin", "phone", "openid", "profile", "email"));
```
raphkim, line 495: are these values stored as enum somewhere? I feel like it would be better to use well-defined enum here if we have any defined, and if not we should create them.

aws-android-sdk-mobile-client/src/androidTest/java/com/amazonaws/mobile/client/AWSMobileClientTest.java
```diff
@@ -313,7 +313,7 @@ public void testSignUp() throws Exception {
```
raphkim, line 313:
```
try {
    UUID.fromString(id);
} catch (IllegalArgumentException error) {
    fail("userSub must be a valid UUID");
}
```

raphkim: oh this was in response to an unrelated section of the code:
> assertNotNull(userSub);
// Validate that the userSub is a valid UUID
        assertEquals(36, userSub.length());
        assertEquals(5, userSub.split("-").length);
        assertEquals(8,  userSub.split("-")[0].length());
        assertEquals(4,  userSub.split("-")[1].length());
        assertEquals(4,  userSub.split("-")[2].length());
        assertEquals(4,  userSub.split("-")[3].length());
        assertEquals(12,  userSub.split("-")[4].length());

raphkim: EDIT: nevermind! This suggestion is not a good solution according to this [stack overflow](https://stackoverflow.com/questions/20041051/how-to-judge-a-string-is-uuid-type) discussion! Another way to test would be 
```
assertEquals(userSub, UUID.fromString(userSub).toString());
```

aws-android-sdk-mobile-client/src/androidTest/java/com/amazonaws/mobile/client/AWSMobileClientCustomAuthTest.java
TrekSoft: This is the specific answer setup in the Lambda function.

aws-android-sdk-mobile-client/src/androidTest/java/com/amazonaws/mobile/client/AWSMobileClientPersistenceWithRestartabilityTest.java
TrekSoft: These are specific strings from the awsconfiguration file so I don't think an ENUM would make sense in this case.

aws-android-sdk-mobile-client/src/androidTest/java/com/amazonaws/mobile/client/AWSMobileClientTest.java
TrekSoft: Got it - thanks for the suggestion but I think we should leave existing code as is unless it's broken. :)

PR #1990, opened 2020-07-07
feat: migrate to android x

jamesonwilliams:
__Adopts [Android X](https://developer.android.com/jetpack/androidx) dependencies__.

This is a prerequisite, to be able to update the SDK's various dependencies towards secure, modern, versions. Android X is a dependency of most recent package versions.

This is a breaking change and will go into the 2.17.0 release.

__Some specific consequences__:
 - The project `minSdk` becomes _at least_ 15. Certain modules may continue to cite a higher `minSdk` for module-specific reasons.
 - This change is "viral". Consumers will need to use Android X in their projects as well.

__Some specific changes__:

 - `android.support.*` imports become `andoridx.*`;
 - `InstrumentationRegistry` is replaced by `ApplicationProvider` in test;
 - `AndoridJUnit4Runner` is no longer used in test;
 - The following lines are applied in `gradle.properties`: 
    ```gradle.properties
    android.useAndroidX=true
    android.enableJetifier=true
    ```


gradle.properties
```diff
@@ -26,4 +26,8 @@ POM_LICENCE_DIST=repo
 
 POM_DEVELOPER_ID=amazonwebservices
 POM_DEVELOPER_ORGANIZATION=Amazon Web Services
-POM_DEVELOPER_ORGANIZATION_URL=http://aws.amazon.com
\ No newline at end of file
+POM_DEVELOPER_ORGANIZATION_URL=http://aws.amazon.com
+
+android.useAndroidX=true
+android.enableJetifier=true
```
arturdryomov, line 32: If no dependencies use support-era dependencies — I suggest to remove this line. No need to run Jetifier on artifacts if not necessary.

jamesonwilliams: > If no dependencies use support-era dependencies — I suggest to remove this line. No need to run Jetifier on artifacts if not necessary.

Thanks for the callout. After this #1990 goes in, I'm going to come back through and update a bunch of dependencies, to their latest stable release versions. I will re-evaluate the need for `android.enableJetifier=true` once I have completed the dependencies updates.

aws-android-sdk-cognitoidentityprovider-test/build.gradle
```diff
@@ -2,14 +2,14 @@ apply plugin: 'com.android.library'
...
+    compileSdkVersion 28
 
     defaultConfig {
         minSdkVersion 15
-        targetSdkVersion 27
+        targetSdkVersion 28
         versionCode 1
         versionName '1.0'
-        testInstrumentationRunner 'android.support.test.runner.AndroidJUnitRunner'
+        testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
```
richardmcclellan, line 12: According to your PR description, `AndroidJUnit4Runner` is no longer used, so can this be removed?  (if so, remove it from all the build.gradle files in this PR)

jamesonwilliams: They are similarly named. This is the AndroidJUnit___Runner___; the old one was the AndroidJUnit__4Runner__. It can be selected by stating a [`testInstrumentationRunner`](https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.DefaultConfig.html#com.android.build.gradle.internal.dsl.DefaultConfig:testInstrumentationRunner) in Gradle, whereas the old one with the "4" in its name had to be elected in an `@RunWith` annotation.

PR #1993, opened 2020-07-07
feat: run integ tests on feature branches too

jamesonwilliams:


jamesonwilliams:
> Fundamentally, running integration tests on feature branches is definitely the right way to go.
> 
> However, this could increase the cost of CircleCI. Is that a concern at all? I believe CircleCI usually provides a pool of containers for a fixed price, and when the pool is full, requests are queued, so I suppose it won't actually increase cost, but could make tests take longer if we are hitting capacity more often.

Thanks for raising this topic @richardmcclellan. Let's keep our eyes peeled to make sure things aren't getting queued up too much. I'd wager that -- at the moment -- our PR traffic on this repo is probably low enough that queuing is a theoretical, and not a pressing, matter of concern.

PR #1994, opened 2020-07-07
Disable unused Gradle Android plugin features

arturdryomov:


gradle.properties
```diff
@@ -12,6 +12,14 @@ org.gradle.jvmargs=-Xmx1536m
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 
+# Disable Gradle Android plugin features.
+# Reference: https://developer.android.com/studio/releases/gradle-plugin#4-0-0-new
+android.defaults.buildfeatures.buildconfig=true
```
arturdryomov, line 17: This is the most important line. There is no need to ship the generated `BuildConfig` to consumers since they have their own ones which reflect application version, package and user fields better.

The remaining is a bonus of disabling irrelevant tasks which take non-zero time to execute.

arturdryomov:
PTAL @jamesonwilliams

arturdryomov:
As far as I see, the `aws-android-sdk-auth-ui` depends on the AppCompat which has neither. From experiments, the AGP will fail loudly in compile time when a disabled feature is used. For example, attempting to add a `BuildConfig` field will throw an error with a message that this feature is disabled.

arturdryomov:
@jamesonwilliams, is there anything I can help with to get this one merged?

jamesonwilliams:
> @jamesonwilliams, is there anything I can help with to get this one merged?

Hey @arturdryomov! I had a question above, about the true/false values of the options. The default values for these options are:
1. [`buildconfig` is true](https://developer.android.com/reference/tools/gradle-api/com/android/build/api/dsl/BuildFeatures),
2. [`aidl` is true](https://developer.android.com/reference/tools/gradle-api/com/android/build/api/dsl/BuildFeatures#aidl:kotlin.Boolean)
3. [`renderScript` is true](https://developer.android.com/reference/tools/gradle-api/com/android/build/api/dsl/BuildFeatures#renderScript:kotlin.Boolean)
4. [`resValues` is true](https://developer.android.com/reference/tools/gradle-api/com/android/build/api/dsl/BuildFeatures#resValues:kotlin.Boolean)
5. [`shaders` is true](https://developer.android.com/reference/tools/gradle-api/com/android/build/api/dsl/BuildFeatures#shaders:kotlin.Boolean)

TLDR: I think you meant to set all of these to ___false___, to mean "don't use this functionality," right?

arturdryomov:
OMG, I’m an idiot. Switched values, thanks! 

That’s like the third time I’m doing this kind of change with `true` instead of `false`.

jamesonwilliams:
> OMG, I’m an idiot. Switched values, thanks!
> 
> That’s like the third time I’m doing this kind of change with `true` instead of `false`.

No worries. The simplest mistakes are the easiest to overlook! That's why we do code review, I guess. :-D

PR #1997, opened 2020-07-08
chore: use latest dependency versions

jamesonwilliams:
The SDK has not been receiving regular updates to its dependencies. As a result, many of them have drifted out of date. This large "catch up" commit
updates all dependencies to the latest versions available from Maven.

This change may have impact to the consumer's dependency graph, and may be
considered a breaking change. It will go into the 2.17.0 release.

-------------

The `build.gradle` and `pom.xml` of each module is updated to include the latest
versions available through Maven.

The `pom.xml`s had drifted from the `build.gradle`s, so the change made here is not a directly matching update. After this commit, the `pom.xml`s should accurately reflect the current state of the
`build.gradle` dependencies. Future work will onboard a plugin to automatically
generate the `pom.xml`, to better ensure they are kept in sync.

### DIFF
--- a/aws-android-sdk-auth-google/build.gradle
+++ b/aws-android-sdk-auth-google/build.gradle
@@ -17,9 +17,8 @@ android {
 
 dependencies {
     api project(':aws-android-sdk-auth-core')
-    api 'com.google.android.gms:play-services-auth:9.8.0'
-    implementation 'androidx.legacy:legacy-support-v4:1.0.0'
-
-    testImplementation 'junit:junit:4.12'
+    api 'com.google.android.gms:play-services-auth:18.0.0'
+    implementation 'androidx.annotation:annotation:1.1.0'
+    implementation 'androidx.appcompat:appcompat:1.1.0'
 }
 
--- a/aws-android-sdk-core/build.gradle
+++ b/aws-android-sdk-core/build.gradle
@@ -9,7 +9,6 @@ android {
         targetSdkVersion 28
         versionCode 1
         versionName '1.0'
-        testInstrumentationRunner 'androidx.test.runner.AndroidJUnitRunner'
 
         consumerProguardFiles 'consumer-proguard-rules.pro'
     }
@@ -20,14 +19,14 @@ android {
 }
 
 dependencies {
-    api 'com.google.code.gson:gson:2.2.4'
-    implementation 'com.fasterxml.jackson.core:jackson-core:2.1.1'
+    api 'com.google.code.gson:gson:2.8.6'
+    implementation 'com.fasterxml.jackson.core:jackson-core:2.11.0'
 
-    testImplementation 'joda-time:joda-time:2.0'
-    testImplementation 'junit:junit:4.12'
+    testImplementation 'joda-time:joda-time:2.8.1'
+    testImplementation 'junit:junit:4.13'
     testImplementation 'org.apache.commons:commons-io:1.3.2'
     testImplementation 'org.easymock:easymock:3.1'
-    testImplementation 'org.robolectric:robolectric:3.8'
+    testImplementation 'org.robolectric:robolectric:4.3.1'
     testImplementation 'xerces:xercesImpl:2.12.0'
     testRuntimeOnly 'org.apache.httpcomponents:httpclient:4.5.12'
 }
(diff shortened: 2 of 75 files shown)


aws-android-sdk-auth-google/build.gradle
```diff
@@ -17,9 +17,8 @@ android {
 
 dependencies {
     api project(':aws-android-sdk-auth-core')
-    api 'com.google.android.gms:play-services-auth:9.8.0'
-    implementation 'androidx.legacy:legacy-support-v4:1.0.0'
-
-    testImplementation 'junit:junit:4.12'
+    api 'com.google.android.gms:play-services-auth:18.0.0'
```
raphkim, line 20: this looks like a huge version jump. this didn't break anything?

aws-android-sdk-core/build.gradle
```diff
@@ -20,14 +19,14 @@ android {
 }
 
 dependencies {
-    api 'com.google.code.gson:gson:2.2.4'
-    implementation 'com.fasterxml.jackson.core:jackson-core:2.1.1'
+    api 'com.google.code.gson:gson:2.8.6'
+    implementation 'com.fasterxml.jackson.core:jackson-core:2.11.0'
```
raphkim, line 23: ah I see what you meant hahaha. Where do we use jackson even?

aws-android-sdk-core/build.gradle
```diff
@@ -20,14 +19,14 @@ android {
...
 
-    testImplementation 'joda-time:joda-time:2.0'
-    testImplementation 'junit:junit:4.12'
+    testImplementation 'joda-time:joda-time:2.8.1'
+    testImplementation 'junit:junit:4.13'
     testImplementation 'org.apache.commons:commons-io:1.3.2'
     testImplementation 'org.easymock:easymock:3.1'
-    testImplementation 'org.robolectric:robolectric:3.8'
+    testImplementation 'org.robolectric:robolectric:4.3.1'
```
raphkim, line 29: this may break some things. I distinctly remember having to set the version to 3.8 because of some incompatibilities in our code. Can you confirm that this is safe?

aws-android-sdk-iot/build.gradle
```diff
@@ -12,14 +12,14 @@ android {
...
 
 dependencies {
-    api (project(':aws-android-sdk-core')){
+    api (project(':aws-android-sdk-core')) {
         exclude group: 'com.google.android', module: 'android'
     }
-    implementation 'org.conscrypt:conscrypt-android:2.0.0'
-    implementation 'org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.2'
+    implementation 'org.conscrypt:conscrypt-android:2.4.0'
+    implementation 'org.eclipse.paho:org.eclipse.paho.client.mqttv3:1.2.4'
```
raphkim, line 19: Didn't the paho update from 1.2.2 to 1.2.3 break something and we had to reverse?

aws-android-sdk-core/build.gradle
raphkim: We should also confirm with other robolectric dependencies throughout the repo

aws-android-sdk-core/build.gradle
jamesonwilliams: No clue. After this goes in, I'm going to come back and see if I can remove one of these.

aws-android-sdk-core/build.gradle
jamesonwilliams: The confirmation is that the unit tests are passing in the CI/CD! :-D

aws-android-sdk-iot/build.gradle
jamesonwilliams: 1.2.3 did break. 1.2.4 appears to be fixed again. I ran the IoT tests.

aws-android-sdk-auth-google/build.gradle
jamesonwilliams: Not at compile-time, anyway. We don't have good test coverage on this module, so it's difficult to know what if any runtime issues may have been resolved / created.

aws-android-sdk-core/build.gradle
raphkim: It's possible that the issues were with the integration tests that are now removed. It sounds fine in that case.

PR #1998, opened 2020-07-08
chore: target latest api level 29

jamesonwilliams:
Changes `targetSdkVersion` and `compileSdkVersion` to the latest stable SDK
version, 29.

### DIFF
--- a/aws-android-sdk-apigateway-core/build.gradle
+++ b/aws-android-sdk-apigateway-core/build.gradle
@@ -1,11 +1,11 @@
 apply plugin: 'com.android.library'
 
 android {
-    compileSdkVersion 28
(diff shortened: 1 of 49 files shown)


jamesonwilliams:
I'm hesitant to include this right now. 

The build now includes these warnings:
```
warning: /opt/android/sdk/platforms/android-30/optional/org.apache.http.legacy.jar(org/apache/http/HttpStatus.class): major version 53 is newer than 52, the highest major version supported by this compiler.
  It is recommended that the compiler be upgraded.
```

It may make sense for us to fully migrate off of the Apache HTTP client before proceeding. The support jar is likely a dead end path, and it is currently causing some compatibility concerns with this warning.

PR #2002, opened 2020-07-09
aws-android-sdk-secretsmanager Update models to latest

awsmobilesdk:
aws-android-sdk-secretsmanager Update models to latest


phani-srikar:
This is an unsupported model but we had discussions at some point, if we want to support this.

jamesonwilliams:
> This is an unsupported model but we had discussions at some point, if we want to support this. If we still want to support this, please check if it builds successfully before merging it in. Otherwise, close the PR and I will remove this from supported models catalog.

Hey @phani-srikar. IIRC, we had discussed the possibility of using the Secrets Manager in our test automation project, [here](https://github.com/aws-amplify/amplify-ci-support). Ultimately we used the SSM Parameter Store, instead.

Secrets Manager sounds like a reasonable thing to include in the SDK. At the same time, I am currently trying to minimize the scope of things supported by the mobile SDK, so that we can do a better job with the the things that are active.

My suggestion, for the time-being, is that we close this PR and wait for customers to request its inclusion. (Or, for a convincing use case in our own test infrastructure.) We could include it at that time.

PR #2004, opened 2020-07-09
CHANGELOG for model updates published around: Thu Jul 09 18:09:04 UTC 2020

awsmobilesdk:
CHANGELOG for model updates published around: Thu Jul 09 18:09:04 UTC 2020


jamesonwilliams:
We will have to hand-author a changelog entry. It will mention SNS and Comprehend, but not Secrets Manager.

PR #2006, opened 2020-07-10
Change Gradle Android SDK values to be reuseable

arturdryomov:
Dependencies can be organized this way as well but doing both in a single change feels too dangerous.

PTAL @jamesonwilliams

### DIFF
--- a/buildSrc/src/main/java/AndroidSdk.java
+++ b/buildSrc/src/main/java/AndroidSdk.java
@@ -1,6 +1,6 @@
 class AndroidSdk {
-		public static final int COMPILE = 28;
-		public static final int MINIMUM = 15;
-		public static final int MINIMUM_TEST = 18; // UI Automator
-		public static final int TARGET = 28;
+    public static final int COMPILE = 28;
+    public static final int MINIMUM = 15;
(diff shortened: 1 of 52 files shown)


buildSrc/src/main/java/AndroidSdk.java
```diff
@@ -0,0 +1,6 @@
+class AndroidSdk {
+    public static final int COMPILE = 28;
+    public static final int MINIMUM = 15;
```
jamesonwilliams, line 3: 15 isn't actually a global minimum right now -- some of the `build.gradle` sort of un-intentionally lie. The various UI modules use a range of different API features. 21 or 23 is probably the actual minimum SDK for some of them.

Also: one of my next steps for the SDK build system was to consolidate all of these module-specific files into the central root-level `build.gradle` as we do [in Amplify](https://github.com/aws-amplify/amplify-android/blob/main/build.gradle#L96-L145).

How do you see this PR fitting in with these two considerations?

arturdryomov: Huh, interesting. Thanks for the clarification! A single configuration point certainly sounds better than a magnitude (even with reused values).

PR #2007, opened 2020-07-11
fix: derrive min sdk values from android lint

jamesonwilliams:
For the purpose of re-enabling NewApi and other API-related Android Lint
warnings, all `abortOnError` options are removed. In response, some
unrelated lint warnings are being addressed or suppressed in code, as a
result of this commit.

To derive an appropriate `minSdkVersion` for each module, `minSdkVersion`
was stripped from all modules, and then updated according to the project
build requirements.

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/internal/keyvaluestore/AWSKeyValueStore.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/internal/keyvaluestore/AWSKeyValueStore.java
@@ -61,10 +61,6 @@ public class AWSKeyValueStore {
     // SecureRandom is expensive.
     private SecureRandom secureRandom;
 
-    private static final int ANDROID_API_LEVEL_23 = 23;
-    private static final int ANDROID_API_LEVEL_18 = 18;
-    private static final int ANDROID_API_LEVEL_10 = 10;
-
     private static final String CIPHER_AES_GCM_NOPADDING = "AES/GCM/NoPadding";
     private static final int CIPHER_AES_GCM_NOPADDING_IV_LENGTH_IN_BYTES = 12;
     private static final int CIPHER_AES_GCM_NOPADDING_TAG_LENGTH_LENGTH_IN_BITS = 128;
@@ -102,8 +98,6 @@ public class AWSKeyValueStore {
 
     private static final int AWS_KEY_VALUE_STORE_VERSION = 1;
 
-    private int apiLevel;
-
     private static Map<String, String> getCacheForKey(String key) {
         if (cacheFactory.containsKey(key)) {
             return cacheFactory.get(key);
@@ -129,7 +123,6 @@ public class AWSKeyValueStore {
         this.secureRandom = new SecureRandom();
         this.cache = getCacheForKey(sharedPreferencesName);
         this.sharedPreferencesName = sharedPreferencesName;
-        this.apiLevel = Build.VERSION.SDK_INT;
         this.context = context;
         setPersistenceEnabled(isPersistenceEnabled);
     }
@@ -156,7 +149,7 @@ public class AWSKeyValueStore {
 
                 initKeyProviderBasedOnAPILevel();
 
-                logger.info("Detected Android API Level = " + apiLevel);
+                logger.info("Detected Android API Level = " + Build.VERSION.SDK_INT);
                 logger.info("Creating the AWSKeyValueStore with key for " +
                         "sharedPreferencesForData = " + sharedPreferencesName);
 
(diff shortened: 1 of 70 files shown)


aws-android-sdk-core/src/main/java/com/amazonaws/internal/keyvaluestore/AWSKeyValueStore.java
```diff
@@ -471,11 +464,11 @@ private AlgorithmParameterSpec getInitializationVector(final String keyOfDataInS
     }
 
     private AlgorithmParameterSpec getAlgorithmParameterSpecForIV(byte[] iv) {
-        return
-            //@apiLevel23Start
-            apiLevel >= ANDROID_API_LEVEL_23 ? new GCMParameterSpec(CIPHER_AES_GCM_NOPADDING_TAG_LENGTH_LENGTH_IN_BITS, iv) :
-            //@apiLevel23End
-            new IvParameterSpec(iv);
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
jamesonwilliams, line 467: Android Lint can't figure out that this code is correctly guarded, unless these standard `Build` values are used.

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/encoding/ChunkDecoder.java
```diff
@@ -53,7 +53,7 @@ private ChunkDecoder() {
 
         try {
             final String headersAsString =
-                    readInputStream(inputStream, PAYLOAD_DELIMITER.getBytes(StandardCharsets.UTF_8));
```
jamesonwilliams, line 56: `StandardCharsets` bumps the `minSdkVersion` up to 19, for no good reason. Don't use it.

aws-android-sdk-core/src/main/java/com/amazonaws/internal/keyvaluestore/KeyNotGeneratedException.java
```diff
@@ -16,8 +16,5 @@ public KeyNotGeneratedException(String s, Throwable throwable) {
     public KeyNotGeneratedException(Throwable throwable) {
         super(throwable);
     }
-
-    protected KeyNotGeneratedException(String s, Throwable throwable, boolean b, boolean b1) {
-        super(s, throwable, b, b1);
```
jamesonwilliams, line 21: This was not being used, and it has an insanely high API level requirement. Get rid of it!

aws-android-sdk-kinesisvideo-archivedmedia/build.gradle
```diff
@@ -4,16 +4,12 @@ android {
     compileSdkVersion 28
 
     defaultConfig {
-        minSdkVersion 15
+        minSdkVersion 21 // for kinesisvideo, which uses camera2 APIs.
```
TrekSoft, line 7: Hmm, will this have an impact on our existing customers? What was happening currently with the minSdkVersion being too low? Is there only a limited number of methods needing this we could wrap with RequiresAPI annotations?

aws-android-sdk-kinesisvideo/build.gradle
```diff
@@ -5,7 +5,7 @@ android {
     useLibrary 'org.apache.http.legacy'
 
     defaultConfig {
-        minSdkVersion 15
+        minSdkVersion 21 // android.hardware.camera2
```
TrekSoft, line 8: Same question here as before

aws-android-sdk-mobile-client/src/main/AndroidManifest.xml
```diff
@@ -1,9 +1,17 @@
...
            xmlns:amazon="http://schemas.amazon.com/apk/res/android"
+           xmlns:tools="http://schemas.android.com/tools"
            package="com.amazonaws.mobile.client">
 
     <uses-permission android:name="android.permission.INTERNET"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
- 
+
+    <!--
+         The hostest UI requires API 16, for Custom Tabs support in the
```
TrekSoft, line 11: *hosted UI

aws-android-sdk-kinesisvideo-archivedmedia/build.gradle
jamesonwilliams: Previously, the `lintOptions { abortOnError false }` was suppressing the build error about the `minSdkVersion` being too low.

The API level is ultimately coming from a few files in the `-kinesisvideo` package, which then bump up this package. Let me see if I can use the `RequiresVersion`, to lower requirements on the two packages.

aws-android-sdk-mobile-client/src/main/AndroidManifest.xml
jamesonwilliams: Haha! Oops. Will fix.

aws-android-sdk-kinesisvideo-archivedmedia/build.gradle
jamesonwilliams: I started going down this path -- it is ultimately a lot of files that raise that API level. I think it'd be best to do it globally on the module for now. If people complain and need lower API level support, we could re-evaluate in the future.

PR #2008, opened 2020-07-12
fix: remove unneeded test usage of apache http

jamesonwilliams:
The Apache HTTP client was being minimally used as a String processing
utility, not as a network client. In an effort to minimize the code
base's coupling to Apache libraries, this usage is replaced with an line
method that will achieve the same results.

### DIFF
--- a/aws-android-sdk-polly/src/test/java/com/amazonaws/services/polly/AmazonPollyPresigningClientTest.java
+++ b/aws-android-sdk-polly/src/test/java/com/amazonaws/services/polly/AmazonPollyPresigningClientTest.java
@@ -19,6 +19,8 @@ import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
+import android.net.UrlQuerySanitizer;
+
 import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.amazonaws.regions.Region;
@@ -59,6 +61,7 @@ public class AmazonPollyPresigningClientTest {
     private static final String SECRET_KEY = "secret-key";
 
     private static final String INPUT_TEXT = "<speak>Mary has a little lamb</speak>";
+    private static final String SANITIZED_TEXT = "_speak_Mary_has_a_little_lamb_/speak_";
     private static final String SSML_TEXT_TYPE = TextType.Ssml.toString();
     private static final String SALLI_VOICE_ID = VoiceId.Salli.toString();
     private static final String MP3_OUTPUT_FORMAT = OutputFormat.Mp3.toString();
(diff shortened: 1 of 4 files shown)


aws-android-sdk-polly/src/test/java/com/amazonaws/services/polly/AmazonPollyPresigningClientTest.java
```diff
@@ -216,4 +213,39 @@ private void assertBasicUriValues(URI uri) {
...
+                if (pair != null && pair.length() > 0) {
+                    String[] parts = pair.split("=");
+                    list.add(new Pair(parts[0], parts[1]));
+                }
+            }
+            return list;
+        }
+
+        private static final class Pair {
+            private final android.util.Pair<String, String> androidPair;
```
richardmcclellan, line 236: Neat!  I didn't know about `android.util.Pair`.

aws-android-sdk-polly/src/test/java/com/amazonaws/services/polly/AmazonPollyPresigningClientTest.java
```diff
@@ -216,4 +213,39 @@ private void assertBasicUriValues(URI uri) {
...
+        static List<Pair> parse(URI uri) {
+            List<Pair> list = new ArrayList<>();
+            String query = uri.getQuery();
+            if (query == null) {
+                return Collections.emptyList();
+            }
+            for (String pair : query.split("&")) {
+                if (pair != null && pair.length() > 0) {
+                    String[] parts = pair.split("=");
+                    list.add(new Pair(parts[0], parts[1]));
```
richardmcclellan, line 229: I think you need to UTF-8 decode the `key` and `value` here?   

Also, should we be a little more defensive and make sure parts.length == 2?

jamesonwilliams: Resolution: I went back to using the `UrlQuerySanitizer`, so this method has been removed.

aws-android-sdk-polly/src/test/java/com/amazonaws/services/polly/AmazonPollyPresigningClientTest.java
jamesonwilliams: Yea! There are [a few useful things under `android.util.*`](https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-master-dev/core/core/src/main/java/androidx/core/util?source=post_page---------------------------%2F%2F%2F%2F%2F%2F%2F&autodive=0%2F%2F%2F%2F%2F%2F%2F%2F%2F%2F%2F/), and [android.util](https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/util/), it turns out. I haven't looked at many of them.

PR #1981, opened 2020-07-02
fix: remove patently false information from readme

jamesonwilliams:
### [See a rendering of the updated README.md](https://github.com/aws-amplify/aws-sdk-android/blob/0cb5eb9b96f586eb4cd3ebabe215886286f4f76a/README.md#aws-sdk-for-android).

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2013, opened 2020-07-12
Catchup develop with main

jamesonwilliams:
`main` contains changes that aren't in `develop`, but `develop` is supposed to ahead of `main`. To resolve this, merge `origin/main` into `develop`.


jamesonwilliams:
> #1981

Yea, that was my bad on the README PR.

I agree, I think we can probably change the default PR branch... let's discuss offline.

Issue #476, opened 2018-06-08
NoClassDefFoundError: org/apache/commons/logging/LogFactory


emumair:
* What service are you using?
**Mobile Analytics**

* In what version of SDK are you facing the problem?
`implementation 'com.amazonaws:aws-android-sdk-core:2.4.2'`
`implementation 'com.amazonaws:aws-android-sdk-mobileanalytics:2.4.2'`

* Is the issue limited to Simulators / Actual Devices?
**No**

* Can your problem be resolved if you bump to a higher version of SDK?
**No**

* Is this problem related to specific Android/OS version?
**Yes, API level 28**

Runs well on version <= 27 but throws exception when app is run on version 28.

Crash occurs at:

`CognitoCachingCredentialsProvider credentialsProvider = new CognitoCachingCredentialsProvider(context, BuildConfig.AMAZON_IDENTITY_POOL_ID, region);`

* **Stacktrace**

```
06-07 15:58:10.189 5075-5075 E/AndroidRuntime: FATAL EXCEPTION: main
    java.lang.NoClassDefFoundError: Failed resolution of: Lorg/apache/commons/logging/LogFactory;
        at com.amazonaws.util.VersionInfoUtils.<clinit>(VersionInfoUtils.java:41)
        at com.amazonaws.util.VersionInfoUtils.getVersion(VersionInfoUtils.java:54)
        at com.amazonaws.auth.CognitoCachingCredentialsProvider.<clinit>(CognitoCachingCredentialsProvider.java:81)
        at com.umair.myapp.analytics.AmazonMobileAnalytics.initialize(AmazonMobileAnalytics.java:41)
        at com.umair.myapp.MyApplication.onCreate(MyApplication.java:48)
        at android.app.Instrumentation.callApplicationOnCreate(Instrumentation.java:1154)
        at android.app.ActivityThread.handleBindApplication(ActivityThread.java:5858)
        at android.app.ActivityThread.access$1000(ActivityThread.java:197)
        at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1634)
        at android.os.Handler.dispatchMessage(Handler.java:106)
        at android.os.Looper.loop(Looper.java:193)
        at android.app.ActivityThread.main(ActivityThread.java:6642)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:493)
[...]

PR #2014, opened 2020-07-14
chore: remove legacy apache http from core

jamesonwilliams:
The Android SDK uses HttpURLConnection as its HTTP runtime.
aws-android-sdk-core contains dead code that references the legacy
Apache HTTP Client used by the V1 Java SDK, from which this codebase
originates. The Apache HTTP client has considerable baggage on the
Android platform, and can cause runtime failures. In an effort to fully
remove it from the codebase, it is being removed from core.

As of this commit, Apache Logging is still a compile-time dependency in
core. If it is provided at runtime, it will be used. Otherwise, the SDK
will fallback to logging against the Android Log utility. This behavior
is discussed here: https://github.com/aws-amplify/aws-sdk-android/issues/476


aws-android-sdk-core/src/main/java/com/amazonaws/ClientConfiguration.java
```diff
@@ -752,40 +734,6 @@ public ClientConfiguration withConnectionTimeout(int connectionTimeout) {
         return this;
     }
 
-    /**
-     * Checks if the {@link IdleConnectionReaper} is to be started
-     *
-     * @return if the {@link IdleConnectionReaper} is to be started
-     */
-    public boolean useReaper() {
```
TrekSoft, line 760: Do we have any sense how many people may currently be using these public methods?

jamesonwilliams: It's a little hard to say. I searched around on GitHub. This functionality is also in the V1 Java SDK. I _believe_ that the few bonafide external uses I found were for _that_ SDK.

I can also just `@Deprecated` this method, to be safe. But, I figured it'd be low-impact, even though it's a breaking change. I am planning to bump the release version up to `2.17.x` for the next release, anyway.

TrekSoft: Yeah sounds good - given that research I think that's fair.

PR #2017, opened 2020-07-16
fix: handle arrays in error response

zeynepsu:
*Issue #, if available:*

*Description of changes:*
If the error response has a list of structures, we will start parsing it.

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/util/json/JsonUtils.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/util/json/JsonUtils.java
@@ -142,6 +142,87 @@ public class JsonUtils {
     }
 
     /**
+     * Convenient method to convert a JSON string to a map. Arrays
+     * are not discarded, they are read into the value as a string.
+     * Number and boolean are also stored as string.
+     * Nested arrays are not supported.
+     *
+     * @param in reader
+     * @return a non null, unmodifiable, string to string map
+     */
+    @SuppressWarnings("unchecked")
+    public static Map<String, String> jsonToStringMapWithList(Reader in) {
+        AwsJsonReader reader = getJsonReader(in);
+        try {
+            // in case it's empty
+            if (reader.peek() == null) {
+                return Collections.EMPTY_MAP;
+            }
+            Map<String, String> map = new HashMap<String, String>();
+            reader.beginObject();
+            while (reader.hasNext()) {
+                String key = reader.nextName();
+                if (reader.isContainer()) {
+                    if (AwsJsonToken.BEGIN_ARRAY.equals(reader.peek())) {
+                        final StringWriter out = new StringWriter();
+                        final AwsJsonWriter writer = getJsonWriter(out);
+                        reader.beginArray();
+                        writer.beginArray();
+                        try {
+                            while (!AwsJsonToken.END_ARRAY.equals(reader.peek())) {
+                                AwsJsonToken nextToken = reader.peek();
+                                if (AwsJsonToken.BEGIN_OBJECT.equals(nextToken)){
+                                    reader.beginObject();
+                                    writer.beginObject();
+                                } else if (AwsJsonToken.FIELD_NAME.equals(nextToken)){
+                                    String name = reader.nextName();
+                                    // do not write key if the value is a nested array
+                                    if (!AwsJsonToken.BEGIN_ARRAY.equals(reader.peek())) {
+                                        writer.name(name);
+                                    }
+                                } else if (AwsJsonToken.END_OBJECT.equals(nextToken)){
+                                    reader.endObject();
+                                    writer.endObject();
+                                } else if (AwsJsonToken.END_ARRAY.equals(nextToken)) {
+                                    reader.endArray();
+                                    writer.endArray();
+                                } else if (AwsJsonToken.VALUE_STRING.equals(nextToken)
+                                        || AwsJsonToken.VALUE_NUMBER.equals(nextToken)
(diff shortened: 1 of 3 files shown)


aws-android-sdk-core/src/main/java/com/amazonaws/util/json/JsonUtils.java
```diff
@@ -141,6 +141,87 @@ public static AwsJsonWriter getJsonWriter(Writer out) {
...
+                        writer.endArray();
+                        writer.flush();
+                        writer.close();
+                        map.put(key, out.toString());
+                    } else {
+                        // skip non string or non array nodes
+                        reader.skipValue();
+                    }
+                } else {
+                    map.put(key, reader.nextString());
```
kneekey23, line 214: maybe add a comment here to say if its not an array it falls into this else and just gets mapped like normal

aws-android-sdk-core/src/test/java/com/amazonaws/util/json/JsonUtilsTest.java
```diff
@@ -44,6 +44,24 @@
             + "\"object\":{}"
             + "}";
 
+    private static final String JSON_STRING_WITH_ARRAYS = "{\"string\":\"string\","
```
jamesonwilliams, line 47: If we don't already have it, we should add a facility like Amplify Android's [`Resources.java`](https://github.com/aws-amplify/amplify-android/blob/main/testutils/src/main/java/com/amplifyframework/testutils/Resources.java#L33-L36). It will allow you to keep unmodified json in `src/test/resources/core/json-string-with-arrays.json`, and then read it into test:
```java
String jsonStringWithArrays = 
    Resources.readAsJson(context, "core/json-string-with-arrays.json");
```

aws-android-sdk-core/src/test/java/com/amazonaws/util/json/JsonUtilsTest.java
```diff
@@ -58,6 +76,28 @@ public void testJsonToMap() {
         assertNull("object is ignored", map.get("object"));
     }
 
+    @Test
+    public void testJsonToMapWithList() {
+        Map<String, String> map = JsonUtils.jsonToStringMapWithList(new StringReader(JSON_STRING_WITH_ARRAYS));
+        assertEquals("string value", "string", map.get("string"));
```
jamesonwilliams, line 82: Instead of using all of the individual `assertFoo(...)`s, I recommend using a strategy like:
1. Prepare a `Map` as input.
2. Prepare a JSON string, as below, that you can assert against.
3. Make a single assert call.

```
assertEquals(
    new JSONObject()
        .put("logn value", map.get("..."))
        // etc
        .toString(),
    map
);
```

aws-android-sdk-core/src/main/java/com/amazonaws/util/json/JsonUtils.java
```diff
@@ -141,6 +141,87 @@ public static AwsJsonWriter getJsonWriter(Writer out) {
         return jsonToMap(new StringReader(json));
     }
 
+    /**
+     * Convenient method to convert a JSON string to a map. Arrays
+     * are not discarded, they are read into the value as a string.
```
jamesonwilliams, line 146: Huh. The behavior of `jsonToMap` is to discard arrays? That seems weird. I guess if we update that method, it'd break other things, is that right? 

Instead of squashing the list, and instead of using a `Map`, it feels like we should be recursing down the JSON, and calling the unmarshaller on everything we see, and returning a propertly modeled AWS SDK model class of some kind. Like, `FooServiceSpecificException`, that has a code-generated unmarshaller, that knows how to handle all of the fields in its model?

raphkim: @jamesonwilliams  It seems that a list property inside an exception is unprecedented so far, which suggests that it likely does not break anything. It appears that this code was written with an assumption that an error response will not contain any lists, which worked until now.

aws-android-sdk-core/src/test/java/com/amazonaws/util/json/JsonUtilsTest.java
zeynepsu: I do not really want to take a direct dependency on any Json library here, and we do not have wrappers around JsonObject provided by any libraries. I do not think this warrants an addition of that size at this point, it matches the styling of the code around it.

aws-android-sdk-core/src/test/java/com/amazonaws/util/json/JsonUtilsTest.java
zeynepsu: May I propose that we take that on as a refactoring task separate from this PR? We do not have that facility and I would need more time to look at what it looks like.

jamesonwilliams: I think that's okay, particularly given that line 36 currently uses the same technique that you are.

aws-android-sdk-core/src/test/java/com/amazonaws/util/json/JsonUtilsTest.java
jamesonwilliams: `org.json.JSONObject` is [baked into Android](https://developer.android.com/reference/org/json/JSONObject). [`JsonObject`](https://www.javadoc.io/doc/com.google.code.gson/gson/2.8.5/com/google/gson/JsonObject.html) is part of Gson.

PR #2024, opened 2020-07-22
Adds override for User Agent in AWSMobileClient

TrekSoft:
Currently if the user agent value is set in the client config, it may be appended to other values. This PR adds the ability to set a user agent override value which will ensure that nothing is appended on to it and it is sent as the final user agent.

The reason for this change is that Amplify Auth uses AWSMobileClient which uses underlying User Pool and Identity Pool SDKs. This is how user agents are currently handled:

- Someone uses aws-android-sdk-auth-userpools: User Agent = userpools_UA
- Someone uses aws-android-sdk-cognitoidentityprovider: User Agent = cognitoidentityprovider_UA
- Someone uses AWSMobileClient:
   - Methods using userpools: User Agent = userpools_UA + AMC_string
   - Methods using cognito identity provider: User Agent = cognitoidentityprovider_UA + AMC_string

Until now, someone using Amplify Auth would have the same user agent sent as AWSMobileClient because it uses AWSMobileClient. This is a problem since Amplify Auth usage can't be correctly attributed. The solution requested by the product team is that someone using Amplify Auth has the Amplify user agent sent, completely overriding any of the other user agents the SDKs currently use.

### DIFF
--- a/aws-android-sdk-auth-core/src/main/java/com/amazonaws/mobile/auth/core/IdentityManager.java
+++ b/aws-android-sdk-auth-core/src/main/java/com/amazonaws/mobile/auth/core/IdentityManager.java
@@ -223,9 +223,11 @@ public class IdentityManager {
                            final AWSConfiguration awsConfiguration) {
         this.appContext = context.getApplicationContext();
         this.awsConfiguration = awsConfiguration;
-        this.clientConfiguration = new ClientConfiguration().withUserAgent(awsConfiguration.getUserAgent());
+        this.clientConfiguration = new ClientConfiguration()
+                .withUserAgent(awsConfiguration.getUserAgent())
+                .withUserAgentOverride(awsConfiguration.getUserAgentOverride());
         this.credentialsProviderHolder = new AWSCredentialsProviderHolder();
-        createCredentialsProvider(this.appContext, this.clientConfiguration, awsConfiguration.getUserAgentOverride());
+        createCredentialsProvider(this.appContext, this.clientConfiguration);
         this.awsKeyValueStore = new AWSKeyValueStore(appContext, SHARED_PREF_NAME, isPersistenceEnabled);
     }
 
@@ -255,7 +257,7 @@ public class IdentityManager {
         }
 
         this.credentialsProviderHolder = new AWSCredentialsProviderHolder();
-        createCredentialsProvider(this.appContext, this.clientConfiguration, awsConfiguration.getUserAgentOverride());
+        createCredentialsProvider(this.appContext, this.clientConfiguration);
         this.awsKeyValueStore = new AWSKeyValueStore(appContext, SHARED_PREF_NAME, isPersistenceEnabled);
     }
 
@@ -904,8 +906,7 @@ public class IdentityManager {
      *   only useful for unauthenticated users.
      */
     private void createCredentialsProvider(final Context context,
-                                           final ClientConfiguration clientConfiguration,
-                                           final String userAgentOverride) {
+                                           final ClientConfiguration clientConfiguration) {
         Log.d(LOG_TAG, "Creating the Cognito Caching Credentials Provider "
                 + "with a refreshing Cognito Identity Provider.");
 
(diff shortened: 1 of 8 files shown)


aws-android-sdk-auth-core/src/main/java/com/amazonaws/mobile/auth/core/IdentityManager.java
```diff
@@ -225,7 +225,7 @@ public IdentityManager(final Context context,
         this.awsConfiguration = awsConfiguration;
         this.clientConfiguration = new ClientConfiguration().withUserAgent(awsConfiguration.getUserAgent());
```
jamesonwilliams, line 226: Just put the override into the client config, here.

aws-android-sdk-auth-core/src/main/java/com/amazonaws/mobile/auth/core/IdentityManager.java
```diff
@@ -225,7 +225,7 @@ public IdentityManager(final Context context,
...
         this.credentialsProviderHolder = new AWSCredentialsProviderHolder();
-        createCredentialsProvider(this.appContext, this.clientConfiguration);
+        createCredentialsProvider(this.appContext, this.clientConfiguration, awsConfiguration.getUserAgentOverride());
```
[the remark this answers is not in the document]
TrekSoft: Ah good catch - yeah that's definitely the right way to do it. Will fix here.

aws-android-sdk-auth-core/src/main/java/com/amazonaws/mobile/auth/core/IdentityManager.java
```diff
@@ -223,7 +223,9 @@ public IdentityManager(final Context context,
                            final AWSConfiguration awsConfiguration) {
         this.appContext = context.getApplicationContext();
         this.awsConfiguration = awsConfiguration;
-        this.clientConfiguration = new ClientConfiguration().withUserAgent(awsConfiguration.getUserAgent());
+        this.clientConfiguration = new ClientConfiguration()
+                .withUserAgent(awsConfiguration.getUserAgent())
+                .withUserAgentOverride(awsConfiguration.getUserAgentOverride());
```
jamesonwilliams, line 228: Do we need both? Why not just use `withUserAgent` and have it set a single variable? If not provided, that variable could default to its current `DEFAULT_USER_AGENT` (or whatever.)

If we have two, there are four states I need to consider:
1. What is the behavior when neither is provided?
2. What is the behavior when only the user agent is provided?
3. Versus only the override?
4. Or what if both are provided?

What if either are null?

TrekSoft: We do need both because user agent is simply appended on to default values whereas when userAgentOverride is used it changes the behavior (as you can see in this PR) to overwrite the entire value with whatever is in there.

1) A default value is used
2) User agent is appended on to the default value
3) The override is used as the entire value (without the default)
4) Same as 3

No problem if either are null.

raphkim: @TrekSoft I'm a bit confused by the use-case 2. Why is it necessary to ever append another user agent rather than always overriding?

TrekSoft: This is the current default behavior that AWSMobileClient is doing - don't know why all they designed it that way but here I'm simply leaving it working the way it always did before but with the option for Amplify to add this override so it replaces everything with the value we pass in.

jamesonwilliams: All of the other SDKs in `aws-sdk-android` use the following mechanism to override the `User-Agent` header:
```java
ClientConfiguration clientConfiguration = new ClientConfiguration();
clientConfiguration.setUserAgent(UserAgent.string());
```

I would prefer if we could do the exact same, or something very similar, in the `AWSMobileClient`.

TrekSoft: The problem is, if we just set user agent, the AWS Mobile Client value gets appended to it: https://github.com/aws-amplify/aws-sdk-android/blob/3625c63248e043fb6173f2f09d611a49cd7c6c92/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/internal/oauth2/OAuth2Client.java#L503

Let's call and talk about this.

jamesonwilliams: That's probably okay, isn't it?

jamesonwilliams: (We talked offline, @TrekSoft to list some requirements in PR description.)

TrekSoft: Done

PR #2103, opened 2020-09-09
Configure Apache HttpClient to perform hostname verification

lherman-cs:
Since HostnameVerifier is deprecated on android 23+, I had to adapt the OkHTTPClient hostname verifier to the codebase.

### DIFF
--- a/aws-android-sdk-kinesisvideo/build.gradle
+++ b/aws-android-sdk-kinesisvideo/build.gradle
@@ -6,7 +6,7 @@ android {
     useLibrary 'org.apache.http.legacy'
 
     defaultConfig {
-        minSdkVersion 21 // android.hardware.camera2
+        minSdkVersion 24 // https://developer.android.com/reference/javax/net/ssl/X509ExtendedTrustManager
         targetSdkVersion 29
         versionCode 21800
         versionName '2.18.0'
@@ -24,5 +24,6 @@ dependencies {
     implementation 'androidx.annotation:annotation:1.1.0'
     //noinspection DuplicatePlatformClasses
     compileOnly 'org.apache.httpcomponents:httpclient:4.5.12'
+    testImplementation 'junit:junit:4.13'
 }
 
--- a/aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/DistinguishedNameParser.java
+++ b/aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/DistinguishedNameParser.java
@@ -1,0 +1,392 @@
+package com.amazonaws.kinesisvideo.http;
+
+import javax.security.auth.x500.X500Principal;
+
+/**
+ * A distinguished name (DN) parser. This parser only supports extracting a string value from a DN.
+ * It doesn't support values in the hex-string style.
+ *
+ * Adapted from https://github.com/square/okhttp/blob/okhttp_3.9.x/okhttp/src/main/java/okhttp3/internal/tls/DistinguishedNameParser.java
+ */
+final class DistinguishedNameParser {
+    private final String dn;
+    private final int length;
+    private int pos;
+    private int beg;
+    private int end;
+
+    /** Temporary variable to store positions of the currently parsed item. */
+    private int cur;
+
+    /** Distinguished name characters. */
+    private char[] chars;
+
+    DistinguishedNameParser(X500Principal principal) {
+        // RFC2253 is used to ensure we get attributes in the reverse
+        // order of the underlying ASN.1 encoding, so that the most
+        // significant values of repeated attributes occur first.
+        this.dn = principal.getName(X500Principal.RFC2253);
+        this.length = this.dn.length();
+    }
+
+    // gets next attribute type: (ALPHA 1*keychar) / oid
+    private String nextAT() {
+        // skip preceding space chars, they can present after
+        // comma or semicolon (compatibility with RFC 1779)
+        for (; pos < length && chars[pos] == ' '; pos++) {
+        }
+        if (pos == length) {
+            return null; // reached the end of DN
+        }
+
+        // mark the beginning of attribute type
+        beg = pos;
+
+        // attribute type chars
+        pos++;
+        for (; pos < length && chars[pos] != '=' && chars[pos] != ' '; pos++) {
+            // we don't follow exact BNF syntax here:
+            // accept any char except space and '='
+        }
+        if (pos >= length) {
+            throw new IllegalStateException("Unexpected end of DN: " + dn);
+        }
+
+        // mark the end of attribute type
+        end = pos;
+
+        // skip trailing space chars between attribute type and '='
+        // (compatibility with RFC 1779)
+        if (chars[pos] == ' ') {
+            for (; pos < length && chars[pos] != '=' && chars[pos] == ' '; pos++) {
+            }
+
+            if (chars[pos] != '=' || pos == length) {
+                throw new IllegalStateException("Unexpected end of DN: " + dn);
+            }
+        }
+
+        pos++; //skip '=' char
+
+        // skip space chars between '=' and attribute value
+        // (compatibility with RFC 1779)
+        for (; pos < length && chars[pos] == ' '; pos++) {
+        }
+
+        // in case of oid attribute type skip its prefix: "oid." or "OID."
+        // (compatibility with RFC 1779)
+        if ((end - beg > 4) && (chars[beg + 3] == '.')
+                && (chars[beg] == 'O' || chars[beg] == 'o')
+                && (chars[beg + 1] == 'I' || chars[beg + 1] == 'i')
+                && (chars[beg + 2] == 'D' || chars[beg + 2] == 'd')) {
+            beg += 4;
+        }
+
+        return new String(chars, beg, end - beg);
+    }
+
+    // gets quoted attribute value: QUOTATION *( quotechar / pair ) QUOTATION
+    private String quotedAV() {
+        pos++;
+        beg = pos;
+        end = beg;
+        while (true) {
+
+            if (pos == length) {
+                throw new IllegalStateException("Unexpected end of DN: " + dn);
+            }
+
+            if (chars[pos] == '"') {
+                // enclosing quotation was found
+                pos++;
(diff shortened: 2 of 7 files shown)


aws-android-sdk-kinesisvideo/build.gradle
```diff
@@ -6,7 +6,7 @@ android {
     useLibrary 'org.apache.http.legacy'
 
     defaultConfig {
-        minSdkVersion 21 // android.hardware.camera2
+        minSdkVersion 24 // android.hardware.camera2
```
richardmcclellan, line 9: The minSdkVersion for `android.hardware.camera2` is 21, so this comment no longer makes sense.  Could you update the comment to mention what requires 24?

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/DistinguishedNameParser.java
```diff
@@ -0,0 +1,392 @@
+package com.amazonaws.kinesisvideo.http;
+
+import javax.security.auth.x500.X500Principal;
+
+/**
+ * A distinguished name (DN) parser. This parser only supports extracting a string value from a DN.
+ * It doesn't support values in the hex-string style.
+ *
+ * Adapted from https://github.com/square/okhttp/blob/okhttp_3.9.x/okhttp/src/main/java/okhttp3/internal/tls/DistinguishedNameParser.java
```
richardmcclellan, line 9: Did you consider adding a dependency on okhttp instead of copying this in?

lherman-cs: @richardmcclellan I did try to add it to the dependency list, but android blocked it from being imported. But, probably that was a different version, let me try it again to confirm this.

aws-android-sdk-kinesisvideo/build.gradle
lherman-cs: Yep, makes sense.

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/DistinguishedNameParser.java
lherman-cs: @richardmcclellan it looks like it didn't work because it's an internal package. It doesn't seem that OkHTTPClient exposes this as an external API.

aws-android-sdk-kinesisvideo/build.gradle
lherman-cs: @richardmcclellan updated

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/DistinguishedNameParser.java
richardmcclellan: Got it.  A better solution would probably be to go all in and just use `OkHttp` instead of `Apache`, right?     With `OkHttp`, hostname verification is [enabled by default](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/ssl-socket-factory/).

richardmcclellan: Discussed offline - Migrating to `OkHttp` is ideal long term, but for now we are going with this solution.

aws-android-sdk-kinesisvideo/build.gradle
```diff
@@ -22,7 +22,9 @@ dependencies {
         exclude group: 'com.google.android', module: 'android'
     }
     implementation 'androidx.annotation:annotation:1.1.0'
+    implementation 'com.squareup.okhttp3:okhttp:3.9.0'
```
richardmcclellan, line 25: You decided _not_ to add this for now, right?  If so, can you remove it from the build.gradle?

lherman-cs: Oops, sorry. It was accidentally included.

lherman-cs: Updated

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/HostnameVerifyingX509ExtendedTrustManager.java
```diff
@@ -0,0 +1,234 @@
...
+        this.clientSideHostnameVerificationEnabled = clientSideHostnameVerificationEnabled;
+    }
+
+    private X509ExtendedTrustManager getX509ExtendedTrustManager() {
+        TrustManagerFactory factory;
+        try {
+            factory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            factory.init((KeyStore) null);
+        } catch (NoSuchAlgorithmException nsae) {
+            log.error("Unable to initialize default TrustManagerFactory, using no-op X509ExtendedTrustManager", nsae);
```
richardmcclellan, line 48: Do you know what scenarios would result in the `NoSuchAlgorithmException` or the `KeyStoreException`?   It looks like you essentially disabling hostname verification if either of these excecptions occur, by using a no-op instance, is that right?

hassanctech: You're right, we should instead bail if either of those two exceptions are thrown, basically the `NoSuchAlgorithmException` happens if no registered security provider supports a `TrustManagerFactorySpi` implementation of the specified algorithms (https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/TrustManagerFactory.html#getInstance-java.lang.String-).  And the `KeyStoreException` happens when the init fails for some reason (https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/TrustManagerFactory.html#init-java.security.KeyStore-)

Perhaps in this case we may want to call `Security.getProviders()` and relay that in the error message before bailing so we can track down why the failure happened as well as the string `TrustManagerFactory.getDefaultAlgorithm()`

richardmcclellan: I agree, bailing is a better approach than returning a no op instance.

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/KinesisVideoApacheHttpClient.java
```diff
@@ -86,13 +85,18 @@ public CloseableHttpResponse executeRequest() {
...
-            final SSLConnectionSocketFactory sslSocketFactory = new SSLConnectionSocketFactory(builder.build(),
-                    new String[] {"TLSv1.2"}, // TLS protocol, use 1.2 only
-                    null, // TLS ciphers, use default
-                    new NoOpHostNameVerifier());
+            final SSLContext sslContext = SSLContext.getInstance("TLSv1.2");
+            // Initializes context. Either of the first two parameters may be null in which case
+            // the installed security providers will be searched for the highest priority implementation
+            // of the appropriate factory. Likewise, the secure random parameter may be null in which case the default
+            // implementation will be used.
+            sslContext.init(null, null, null);
```
richardmcclellan, line 94: Shouldn't you be initializing the `SSLContext` here the same way you do it in the `SocketFactory`?    As is, I don't think there will be any hostname verification for this client.

lherman-cs: There will still be a hostname verification as it's set a bit later, SSLIOSessionStrategy. I think either way is an okay approach. I have a manual test that verifies that this part works properly.

hassanctech: I think the same, the second arg is the TrustManagers and specifying null for that seems like it would skip hostname verification.

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/HostnameVerifyingX509ExtendedTrustManager.java
```diff
@@ -0,0 +1,233 @@
...
+            return;
+        }
+
+        log.error("Failed to verify host address: %s", hostAddress);
+        log.error("Failed to verify hostname: %s", hostName);
+        throw new CertificateException("Failed to verify both host address and host name");
+    }
+
+
+    private X509ExtendedTrustManager getNoOpInstance() {
```
richardmcclellan, line 174: This method can be deleted now, right?

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/kinesisvideo/http/HostnameVerifyingX509ExtendedTrustManager.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    private X509ExtendedTrustManager getX509ExtendedTrustManager() {
+        TrustManagerFactory factory;
+        try {
+            factory = TrustManagerFactory.getInstance(TrustManagerFactory.getDefaultAlgorithm());
+            factory.init((KeyStore) null);
+        } catch (NoSuchAlgorithmException nsae) {
+            log.error("Unable to initialize default TrustManagerFactory", nsae);
+            throw new RuntimeException(nsae);
```
richardmcclellan, line 49: I'd suggest including a message as well, in addition to the Throwable.  Probably the same message you are logging as an error would work.  This suggestion applies to all 3 places where you are throwing a `RuntimeException`

richardmcclellan (approved):

hassanctech (approved):

PR #2104, opened 2020-09-12
2.19.0

richardmcclellan:
- Merging everything from develop into main
- Adding CHANGELOG for 2.19.0
- Moved versionName and versionCode from all 50+ module build files to the single project build file, so incrementing version is easier.
- Incrementing version to 2.19.0


aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/services/cognitoidentityprovider/model/transform/StartUserImportJobRequestMarshaller.java
```diff
@@ -22,6 +22,8 @@
 import java.io.StringWriter;
 import java.io.Writer;
 
+import android.text.TextUtils;
```
TrekSoft, line 25: This seems not great that they added an import to TextUtils in all these Marshallers with no apparent use but not a blocker since this is code generated.

TrekSoft (approved):
Cat3u3

PR #2125, opened 2020-09-23
translate Update models to latest

awsmobilesdk:
translate Update models to latest


rjuliano (approved):
Ship it. 

Only change that looked impactful was the change from COMPLETED_WITH_ERRORS to COMPLETED_WITH_ERROR. Upon further review, that was only a documentation fix since the enum item itself appears to be called COMPLETED_WITH_ERROR

PR #2182, opened 2020-10-16
Add upload APIs to transferUtility to support inputStream

changxu0306:
The overloaded upload APIs accepts an `InputStream`, writes it to a temporary `File`, and then pass the `File` to the `TransferUtility` API which takes a file. 

Note that only two `upload` methods are overloaded using a new class called `UploadOptions`, which serves as a builder to hold four optional parameters to pass into `upload` methods.


aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +603,65 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+                while ((bytesRead = inputStream.read(buffer)) != -1) {
+                    outStream.write(buffer, 0, bytesRead);
+                    outStream.flush();
+                }
+                inputStream.close();
+                outStream.close();
+            } catch (Exception e) {
+                file = null;
+            }
+        }
```
raphkim, line 631: You seem to be reusing this code in the other method overload as well. How about using moving this logic elsewhere into a separate method for reuse?

raphkim (changes requested):
I think there are several more overloads of `upload(...)`  methods inside `TransferUtility` class that use `File` as an input. Ideally, we should provide an alternative for every single one of them.

Also, some unit tests and integration tests would be really helpful here to make sure that these methods work as intended.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +603,65 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+    /**
+     * Starts uploading the inputStream to the <b>default</b> bucket, using the given key.
+     *
+     * @param key           The key in the specified bucket by which to store the new object.
+     * @param inputStream   The input stream to upload.
+     * @return A TransferObserver used to track upload progress and state
+     */
+    public TransferObserver upload(String key, InputStream inputStream) throws IOException {
+        // Saves the data as a file in the temporary directory
+        File file = File.createTempFile(UUID.randomUUID().toString(), ".tmp");
```
raphkim, line 615: When do these temporary files get removed? Is it possible to remove them as soon as possible after confirming that the upload is successful (this might be a bit tricky)?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +603,65 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+                int bytesRead;
+                // Keep reading until reaches the end of the stream
+                while ((bytesRead = inputStream.read(buffer)) != -1) {
+                    outStream.write(buffer, 0, bytesRead);
+                    outStream.flush();
+                }
+                inputStream.close();
+                outStream.close();
+            } catch (Exception e) {
+                file = null;
```
raphkim, line 629: I guess `upload(key, file)` doesn't like getting a `null` file and throws an error huh? I think it would be better if you can just catch it sooner and directly throw a more helpful exception for the user here.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +603,65 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+
+    /**
+     * Starts uploading the inputStream to the given bucket, using the given key.
+     *
+     * @param bucket        The name of the bucket to upload the new object to.
+     * @param key           The key in the specified bucket by which to store the new object.
+     * @param inputStream   The input stream to upload.
+     * @return A TransferObserver used to track upload progress and state
+     */
+    public TransferObserver upload(String bucket, String key, InputStream inputStream) throws IOException {
```
raphkim, line 643: looks like `IOException` gets thrown if it fails to create a temp file? Maybe we should just hide this implementation detail and throw a customized error?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +603,65 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+                byte[] buffer = new byte[1024];
+                int bytesRead;
+                // Keep reading until reaches the end of the stream
+                while ((bytesRead = inputStream.read(buffer)) != -1) {
+                    outStream.write(buffer, 0, bytesRead);
+                    outStream.flush();
+                }
+                inputStream.close();
+                outStream.close();
+            } catch (Exception e) {
```
raphkim, line 628: What kind of exceptions can be thrown here? Isn't it just `IOException` that is already declared in the method signature? It may even not be necessary to catch any error here and directly throw it. 

To directly contradict what I just said, I think the ideal solution is to throw a more detailed custom exception that lets the user know that upload was unsuccessful for __ reason.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
richardmcclellan: [File.createTempFile](https://developer.android.com/reference/java/io/File#createTempFile(java.lang.String,%20java.lang.String,%20java.io.File)) automatically appends some UUID or equivalent to ensure the filename is unique, so the prefix doesn't need to be a UUID itself.  Per the docs, it should be at least three characters long, and is recommended that it be a short, meaningful string.  Maybe "awss3"?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +603,65 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+            try {
+                OutputStream outStream = new FileOutputStream(file);
+                byte[] buffer = new byte[1024];
+                int bytesRead;
+                // Keep reading until reaches the end of the stream
+                while ((bytesRead = inputStream.read(buffer)) != -1) {
+                    outStream.write(buffer, 0, bytesRead);
+                    outStream.flush();
+                }
+                inputStream.close();
```
richardmcclellan, line 626: Instead of manually closing the streams at the end, can you use a [`try`-with-resources](https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) here?  That way, it will ensure they are still closed, even if an exception is throw while writing.

raphkim: try-with-resources is the next best thing since sliced bread 😆

aws-android-sdk-s3/build.gradle
```diff
@@ -5,7 +5,7 @@ android {
     compileSdkVersion 29
 
     defaultConfig {
-        minSdkVersion 9
+        minSdkVersion 19
```
changxu0306, line 8: Note that try-with-resources statement requires API level 19.

rjuliano: That's interesting. I think we're supposed to support API 16 and up, so we may have to go without the try-with-resources. Let's discuss with the rest of the team to make sure.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +605,79 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+     *
+     * @param key           The key in the specified bucket by which to store the new object.
+     * @param inputStream   The input stream to upload.
+     */
+    public void uploadInputStream(String key, InputStream inputStream) {
+        try {
+            final File file = writeInputStreamToFile(inputStream);
+            TransferObserver observer = upload(key, file);
+
+            observer.setTransferListener(new TransferListener() {
```
rjuliano, line 619: You probably have one more opportunity to clear up some duplicate code. If you create a private method that creates a TransferListener, I think you'd be in pretty good shape here. Something like this:

```
private TransferListener getListenerFor(File file) {
        return new TransferListener() {
                @Override
                public void onStateChanged(int id, TransferState state) {
                    switch (state) {
                        case COMPLETED:
                            file.delete();
                            break;
                        case FAILED:
                            throw new RuntimeException("Transfer failed.");
                    }
                }

                @Override
                public void onProgressChanged(int id, long bytesCurrent, long bytesTotal) {}

                @Override
                public void onError(int id, Exception ex) {
                    throw new RuntimeException(ex);
                }
            }
    }
```

Then the two call sites would look like:
```
observer.setTransferListener(getListenerFor(file));
```

rjuliano (commented):
This is looking pretty good. I think you do home one more spot you could refactor to get rid of some duplicate code, but you should be in good shape after that.

aws-android-sdk-s3/build.gradle
richardmcclellan: > Note that try-with-resources statement requires API level 19.

I thought this at first as well.  However, it should actually work as long as you are using gradle plugin 3.0.0 or above which we are.  It includes a tool called `desugar`, which adds backwards compatibility for try-with-resources for all Android API versions.  See the reference about halfway down this page about it: https://developer.android.com/studio/write/java8-support

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -762,6 +843,32 @@ private int createMultipartUploadRecords(String bucket, String key, File file, O
...
+        }
+        try (OutputStream outStream = new FileOutputStream(file)) {
+            byte[] buffer = new byte[1024];
+            int bytesRead;
+            // Keep reading until reaches the end of the stream
+            while ((bytesRead = inputStream.read(buffer)) != -1) {
+                outStream.write(buffer, 0, bytesRead);
+                outStream.flush();
+            }
+            inputStream.close();
```
richardmcclellan, line 864: As written, this `inputStream` will only be closed if any of the code above does not throw an exception.    Normally to solve this, you would add a `finally` block after the `catch`, and close it there, so that it gets closed no matter what.  

That said, [whoever opens a stream is typically responsible for closing it](https://stackoverflow.com/a/23925742/282279).  Since the customer will be opening the `InputStream`, I think we should leave it up to the customer to close it.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +605,79 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+                @Override
+                public void onProgressChanged(int id, long bytesCurrent, long bytesTotal) {}
+
+                @Override
+                public void onError(int id, Exception ex) {
+                    throw new RuntimeException(ex);
+                }
+            });
+        } catch (IOException e) {
+            LOGGER.error("Failed writing the input stream into a file: ", e);
```
richardmcclellan, line 640: If all we do is log an error here, then the customer won't have a way to handle it.   I would suggest that we just not catch this error, and let this new upload method throw an IOException for the customer to handle.

richardmcclellan (changes requested):
TransferUtility currently has 10 overloaded implementations of `upload` 😱 😱 😱 
```
1. public TransferObserver upload(String bucket, String key, File file)
2. public TransferObserver upload(String key, File file)
3. public TransferObserver upload(String bucket, String key, File file, CannedAccessControlList cannedAcl)
4. public TransferObserver upload(String key, File file, CannedAccessControlList cannedAcl)
5. public TransferObserver upload(String bucket, String key, File file, ObjectMetadata metadata)    
6. public TransferObserver upload(String key, File file, ObjectMetadata metadata)
7. public TransferObserver upload(String bucket, String key, File file, ObjectMetadata metadata, CannedAccessControlList cannedAcl)
8. public TransferObserver upload(String key, File file, ObjectMetadata metadata, CannedAccessControlList cannedAcl)
9. public TransferObserver upload(String bucket, String key, File file, ObjectMetadata metadata, CannedAccessControlList cannedAcl, TransferListener listener)
10. public TransferObserver upload(String key, File file, ObjectMetadata metadata, CannedAccessControlList cannedAcl, TransferListener listener)
```

These are all  variations that accept up to 6 parameters:
 - `String key` (required)
 - `File file` (required)
 - `String bucket` (optional, defaults to `getDefaultBucketOrThrow()`)
 - `ObjectMetadata metadata` (optional, defaults to new `ObjectMetadata()`)
 - `CannedAccessControlList cannedAcl` (optional, defaults to `null`)
 - `TransferListener listener` (optional, defaults to `null`)


This PR provides `InputStream` versions of upload analogous to 1. & 2.  Amplify [currently calls](https://github.com/aws-amplify/amplify-android/blob/12ec8e9819228f63bd677a0431a5ab191f79d9e1/aws-storage-s3/src/main/java/com/amplifyframework/storage/s3/service/AWSS3StorageService.java#L147) 5., so we at least need to provide an `InputStream` version of that one.    We should also either add `InputStream` support for all 10, or make sure there's a maintainable way of doing so in the future.

We _could_ add 10 more overloads, for a total of 20, but that would be pretty gross and very difficult to maintain.    Instead, I'd suggest creating **two** new methods:

`public TransferObserver upload(String key, InputStream inputStream)`
`public TransferObserver upload(String key, InputStream inputStream, UploadOptions options)`

`UploadOptions` is a new class that can hold all of the optional parameters (bucket, metadata, cannedAcl, transferListener).

Customers could use it like this:
```
transferUtility.upload(key, inputStream, UploadOptions.builder()
        .bucket(bucket)
        .objectMetadata(metadata)
        .cannedAcl(cannedAcl)
        .transferListener(transferListener)
        .build());
```

Or, if you only want to specify a subset of options:
```
transferUtility.upload(key, inputStream, UploadOptions.builder()
        .bucket(bucket)
        .build());
```

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -762,6 +843,32 @@ private int createMultipartUploadRecords(String bucket, String key, File file, O
...
+            int bytesRead;
+            // Keep reading until reaches the end of the stream
+            while ((bytesRead = inputStream.read(buffer)) != -1) {
+                outStream.write(buffer, 0, bytesRead);
+                outStream.flush();
+            }
+            inputStream.close();
+        } catch (IOException e) {
+            file.delete();
+            throw new IOException("Error writing the inputStream into a file.", e);
```
jamesonwilliams, line 867: Single letter variable names are hard to find with search tools. I recommend to avoid the `e`:
```suggestion
        } catch (IOException ioException) {
            file.delete();
            throw new IOException("Error writing the inputStream into a file.", ioException);
```

Other thought: you are catching an `IOException`, only to throw another `IOException`. Maybe you should just let the original exception bubble up, without catching it?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +605,79 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+                @Override
+                public void onProgressChanged(int id, long bytesCurrent, long bytesTotal) {}
+
+                @Override
+                public void onError(int id, Exception ex) {
+                    throw new RuntimeException(ex);
+                }
+            });
+        } catch (IOException e) {
+            LOGGER.error("Failed writing the input stream into a file: ", e);
```
jamesonwilliams, line 677: How will the user know that these errors have a occurred? We need a way to pass these back to the user.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
jamesonwilliams: +1, I think we need to throw the `onError`'s exception on the calling thread, too. Right now, I think it might just crash a background thread, without the consumer having a way to know about it.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
changxu0306: Agree. Since `inputStream` is an input to this method, we should not close it on our will. So I will delete line 864.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
changxu0306: How to do that? Do you mean just remove the catch block? Or just do `throw ioException;` in the catch block?

aws-android-sdk-s3/build.gradle
```diff
@@ -7,10 +7,22 @@ android {
...
+        // Flag to enable support for the new language APIs
+        coreLibraryDesugaringEnabled true
+        // Sets Java compatibility to Java 8
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
     }
 }
 
 dependencies {
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.0.9'
```
changxu0306, line 24: `desugar` is set to make try-with-resources statements compatible with Java 8 without setting minSDK.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,124 @@
...
+    /**
+     * Builder class for UploadOptions
+     */
+    public static class Builder {
+        private String bucket;
+        private ObjectMetadata metadata;
+        private CannedAccessControlList cannedAcl;
+        private TransferListener listener;
+
+        public Builder() {
```
richardmcclellan, line 55: I think this can be private, right?  Since customers will use the static `builder()` below?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,124 @@
+package com.amazonaws.mobileconnectors.s3.transferutility;
+
+import com.amazonaws.services.s3.model.CannedAccessControlList;
+import com.amazonaws.services.s3.model.ObjectMetadata;
+
+/**
+ * A class that can hold all of the optional parameters (bucket, metadata, cannedAcl, transferListener).
+ */
+public class UploadOptions {
+    private String bucket;
```
richardmcclellan, line 10: Make all these `final`

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param bucket        The bucket name
+     * @param metadata      The S3 metadata to associate with this object
+     * @param cannedAcl     The canned ACL to associate with this object
+     * @param listener      The listener to attach to transfer observer
+     */
+    public UploadOptions(String bucket, ObjectMetadata metadata, CannedAccessControlList cannedAcl, TransferListener listener) {
```
richardmcclellan, line 23: Typically with the Builder pattern, this constructor should take a `Builder` object, rather than the list of parameters.   By doing that, this constructor will never grow to have a super long list of parameters that's hard to maintain.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,124 @@
+package com.amazonaws.mobileconnectors.s3.transferutility;
+
+import com.amazonaws.services.s3.model.CannedAccessControlList;
+import com.amazonaws.services.s3.model.ObjectMetadata;
+
+/**
+ * A class that can hold all of the optional parameters (bucket, metadata, cannedAcl, transferListener).
+ */
+public class UploadOptions {
```
richardmcclellan, line 9: Make sure to add `equals`, `hashCode`, and `toString` to this class, in case customers ever do any comparisons with this object.  

These are pretty easy to add with Android Studio - you can type Command + N, and select `equals & hashCode`, and then `toString`.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -762,6 +810,31 @@ private int createMultipartUploadRecords(String bucket, String key, File file, O
...
     }
 
+    private File writeInputStreamToFile(InputStream inputStream) throws IOException {
+        if (inputStream == null) {
+            throw new IllegalArgumentException("Invalid inputStream: " + inputStream);
+        }
+
+        // Saves the data as a file in the temporary directory
+        File file = File.createTempFile("aws-s3", ".tmp");
+        if (!file.isFile()) {
```
richardmcclellan, line 820: Hmm, why would the file not be a file?  Is this check necessary?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +603,48 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+    public TransferObserver upload(String key, InputStream inputStream, UploadOptions options) throws IOException {
+        File file = writeInputStreamToFile(inputStream);
+        TransferObserver observer = upload(
+                options.getBucket() != null ? options.getBucket() : getDefaultBucketOrThrow(),
+                key,
+                file,
+                options.getMetadata() != null ? options.getMetadata() : new ObjectMetadata(), options.getCannedAcl(),
+                options.getTransferListener()
+        );
+        deleteTempFile(observer);
```
richardmcclellan, line 635: We can't delete the `File` until the upload has completed, since the `File` is read during the upload process.  I would suggest putting this code inside the `removeTransferRecordFromDB` method.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
changxu0306: Sorry, it should be `if (!file.isDirectory())`

richardmcclellan: Hmm, in what scenario would it be a directory?  I'm not sure we need to check for that, since we just created it, as a File, not a directory.

changxu0306: I just want to make sure that the file is created successfully. If not, throw an exception. Yeah we shouldn't use isDirectory(). I'm wrong😭

aws-android-sdk-s3/build.gradle
```diff
@@ -7,10 +7,22 @@ android {
...
+        // Required when setting minSdkVersion to 20 or lower
+        multiDexEnabled true
+    }
+
+    compileOptions {
+        // Flag to enable support for the new language APIs
+        coreLibraryDesugaringEnabled true
+        // Sets Java compatibility to Java 8
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
```
jamesonwilliams, line 19: Ahh, right. We have this stuff enabled in Amplify, but I guess it isn't in the `aws-sdk-android`. This seems like a lot to add for one small feature. We might want to just skip the try-with-resources altogether, and just do it the old-fashioned way:
```java
OutputStream outStream = new FileOutputStream(file);
try {
    useTheResource(outStream);
} catch (IOException ioException) {
    // Dang, it didn't work out.
} finally {
    outStream.close();
}
```

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -597,6 +603,48 @@ public TransferObserver upload(String key, File file, ObjectMetadata metadata, C
...
+        return observer;
+    }
+
+    synchronized public void deleteTempFile(TransferObserver observer) {
```
jamesonwilliams, line 639: Doesn't need to be `public`, right?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
richardmcclellan: Got it.  `File.createTempFile` will throw an `IOException` if there's any issue creating it, so that's how you can tell if it was successful.  You aren't catching that exception so it will just be thrown back to the customer, which I think is a reasonable approach.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,124 @@
+package com.amazonaws.mobileconnectors.s3.transferutility;
```
jamesonwilliams, line 1: Needs a copy-right header at the top:
```java
/**
 * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 *
 * Licensed under the Apache License, Version 2.0 (the "License").
 * You may not use this file except in compliance with the License.
 * A copy of the License is located at
 *
 *  http://aws.amazon.com/apache2.0
 *
 * or in the "license" file accompanying this file. This file is distributed
 * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
 * express or implied. See the License for the specific language governing
 * permissions and limitations under the License.
 */
```

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
jamesonwilliams: Ooh I didn't know about the ⌘-N!

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
changxu0306: Make sense. I'll remove this check.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    public TransferListener getTransferListener() {
+        return this.listener;
+    }
+
+    /**
+     * Builder class for UploadOptions
+     */
+    public static class Builder {
```
jamesonwilliams, line 49:
```suggestion
    public static final class Builder {
```

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
jamesonwilliams, line 9:
```suggestion
public final class UploadOptions {
```

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,124 @@
...
+     *                        .cannedAcl(cannedAcl)
+     *                        .transferListener(transferListener)
+     *                        .build());
+     *
+     * @return The builder object to construct a UploadOptions.
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
```
jamesonwilliams, line 123:
```suggestion
```

aws-android-sdk-s3/build.gradle
changxu0306: Got it! 😄

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
raphkim: In fact, we want to hide this detail if possible, so i think it's best left `private`. If a customer happens to have created a file that also begins with "aws-s3" and calls this method not knowing what it does, they will end up losing data.

nit: i think accessor keyword should come first so it should probably be `private synchronized void`

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
raphkim, line 9: should probably add `final` keyword to this too, since we don't want people to extend this class.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,124 @@
...
+        }
+    }
+
+    /**
+     * Minimum calls required.
+     * UploadOptions.builder().bucket(bucket)
+     *                        .objectMetadata(metadata)
+     *                        .cannedAcl(cannedAcl)
+     *                        .transferListener(transferListener)
+     *                        .build());
```
raphkim, line 116: I think there's an extra ")" here

raphkim, line 116: Aren't these optional parameters? If a customer passes in `UploadOptions.builder().build()`, then will it not work?

changxu0306: `UploadOptions.builder().build()` will work of course. I'll remove this javadoc comment if it's confusing.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
changxu0306: If you are using an apple keyboard, it should be 'Control + N' 😄

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,156 @@
...
+        return "UploadOptions{" +
+                "bucket='" + bucket + '\'' +
+                ", metadata=" + metadata +
+                ", cannedAcl=" + cannedAcl +
+                ", listener=" + listener +
+                '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
richardmcclellan, line 142: For `amplify-android` there are a couple checkstyle rules that we follow to ensure a consistent style.  These aren't enforced on `aws-sdk-android`, mostly just because it would be too much effort to go fix all the style issues in the legacy code.  But, for new code like this, we should try to follow the same guidelines.   The ⌘N shortcut gives you a good starting point for `equals`, `hashcode` and `toString`, but some manual modifications are needed.  Here are a few things to address: 
 - While Java does allow you to skip the braces if you only have one statement after an if condition, we always include them anyway, with line breaks before and after, for better readability.
 - Avoid variable names with less than 3 characters.  Short variable names are difficult to search for, when trying to understand code.
 - `Objects.equals` and `Objects.hash` require API 19 and above.  To maintain compatibility with lower API versions, use `ObjectsCompat.equals` and `ObjectsCompat.hash` instead.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferStatusUpdater.java
```diff
@@ -147,6 +148,11 @@ synchronized void removeTransfer(int id) {
     synchronized void removeTransferRecordFromDB(final int id) {
         S3ClientReference.remove(id);
         dbUtil.deleteTransferRecords(id);
+        // Remove temporary file
+        if (dbUtil.getTransferById(id).file.startsWith("aws-s3")) {
+            String path = dbUtil.getRecordUri(id).getPath();
```
richardmcclellan, line 153: I think this is getting a reference to a `Uri` associated with the database entry for this particular upload, not the temporary `File`.  `dbUtil.getTransferById(id).file` should give you the path to the File.

Also, I believe you need to obtain the `File` path before `dbUtil.deleteTransferRecords(id);` is called.   Otherwise, the transfer record won't be there anymore :)

Lastly, `getTransferById(id)` may return null, so it would be good to add a null check after obtaining it.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,156 @@
...
+
+package com.amazonaws.mobileconnectors.s3.transferutility;
+
+import com.amazonaws.services.s3.model.CannedAccessControlList;
+import com.amazonaws.services.s3.model.ObjectMetadata;
+
+import java.util.Objects;
+
+/**
+ * A class that can hold all of the optional parameters (bucket, metadata, cannedAcl, transferListener).
```
richardmcclellan, line 24:
```suggestion
 * Options used to configure the upload of an object using {@link TransferUtility}.
```

richardmcclellan (commented):
I think this is almost there!  The one other thing that would be good to add is unit testing for this new functionality.  The tests should verify that:

 - Calling upload(String key, InputStream inputStream, UploadOptions options) with a bunch of options does in fact call `upload(String bucket, String key, File file, ObjectMetadata metadata, CannedAccessControlList cannedAcl, TransferListener listener)` with all the expected parameters.
 - Verify that the File does in fact get deleted after the upload is complete.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
changxu0306: Got it! That's a good practice for me to learn.😊

changxu0306:
> I think this is almost there! The one other thing that would be good to add is unit testing for this new functionality. The tests should verify that:
> 
>     * Calling upload(String key, InputStream inputStream, UploadOptions options) with a bunch of options does in fact call `upload(String bucket, String key, File file, ObjectMetadata metadata, CannedAccessControlList cannedAcl, TransferListener listener)` with all the expected parameters.
> 
>     * Verify that the File does in fact get deleted after the upload is complete.

The new class `UploadInputStreamTest` performs unit testing to verify those two things and test passed.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferStatusUpdater.java
```diff
@@ -145,6 +146,13 @@ synchronized void removeTransfer(int id) {
      * @param id id of the transfer to remove
      */
     synchronized void removeTransferRecordFromDB(final int id) {
+        // Remove temporary file
+        if (dbUtil.getTransferById(id) != null) {
+            if (dbUtil.getTransferById(id).file.startsWith("aws-s3")) {
```
richardmcclellan, line 151: Can you define "aws-s3" as a constant somewhere that can be referenced by both `TransferStatusUpdater` and `TransferUtility`?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadOptions.java
```diff
@@ -0,0 +1,158 @@
...
+                ", listener=" + listener +
+                '}';
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if (obj == null || getClass() != obj.getClass()) return false;
```
richardmcclellan, line 146:
```suggestion
        if (obj == null || getClass() != obj.getClass()) {
            return false;
        }
```

aws-android-sdk-s3/src/test/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadInputStreamTest.java
```diff
@@ -0,0 +1,221 @@
...
+    public void testDeleteTempFileAfterUpload() throws IOException {
+        String key = getRandomString();
+        final TransferObserver observer = transferUtility.upload(key, inputStream);
+        observer.setTransferListener(new TransferListener() {
+            @Override
+            public void onStateChanged(int id, TransferState state) {
+                Log.d(TAG, "onStateChanged: " + id + ", " + state);
+                switch (state) {
+                    case COMPLETED:
+                        assertFalse(new File(observer.getAbsoluteFilePath()).exists());
```
richardmcclellan, line 182: As written, I believe this test will succeed before `onStateChanged` is called, so this assert statement will never be evaluated.   

// Define a `CountdownLatch` at the top of the test
```
CountDownLatch deleteFileLatch = new CountDownLatch(1);
```

In the callback, call countDown if the file has been deleted:
```
case COMPLETED:
    if(!new File(observer.getAbsoluteFilePath()).exists()) {
        deleteFileLatch.countDown();
    }
```

Then, at the end of the test, validate that the latch counts down within some time period, like 5 seconds:
```
if (!deleteFileLatch.await(2, TimeUnit.SECONDS)) {
    fail("File not deleted after upload.");
}
```

aws-android-sdk-s3/src/test/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadInputStreamTest.java
```diff
@@ -0,0 +1,221 @@
...
+                UploadOptions.builder().bucket(bucket).objectMetadata(metadata).cannedAcl(cannedAcl).build());
+        verify(transferUtility, times(1)).upload(eq(bucket), eq(key), any(File.class),
+                eq(metadata), eq(cannedAcl), isNull(TransferListener.class));
+    }
+
+    @Test
+    public void testUploadWithMetadataAndCannedAclAndListener() throws IOException {
+        String key = getRandomString();
+        transferUtility.upload(key, inputStream,
+                UploadOptions.builder().objectMetadata(metadata).cannedAcl(cannedAcl).transferListener(listener).build());
```
richardmcclellan, line 151: There's not really an issue with the way you have it, but I'd suggest putting each step of the Builder on a new line, as it's just a little bit easier to read that way, and sort of the standard way of using a Builder, like this:

```suggestion
        transferUtility.upload(key, inputStream, UploadOptions.builder()
                .objectMetadata(metadata)
                .cannedAcl(cannedAcl)
                .transferListener(listener)
                .build());
```

Same suggestion for the other places you are using the Builder, in the other tests.

aws-android-sdk-s3/src/test/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadInputStreamTest.java
```diff
@@ -0,0 +1,221 @@
...
+
+    private static final String DEFAULT_BUCKET = "s3-upload-input-stream-unit-test-" + getRandomString();
+    private final InputStream inputStream;
+    private final ObjectMetadata metadata;
+    private final CannedAccessControlList cannedAcl;
+    private final TransferListener listener;
+    private final TransferUtility transferUtility;
+
+    public UploadInputStreamTest() {
+        AmazonS3Client s3 = mock(AmazonS3Client.class);
```
richardmcclellan, line 59: Rather than setting all of this up in a constructor, it's better to put it in a setup method with the @Before annotation, like this:
```
@Before
public void setup() {
    AmazonS3Client s3 = mock(AmazonS3Client.class);
    ....
}
```

With this approach, everything in the `@Before` method will be run before each individual test is run.  This allows you to reset / reinitialize everything to a default state, so that one test won't cause side effects on another test.  This might also fix the `IllegalStateException` that I see most of these tests are throwing right now. 

I'd also suggest creating a teardown method with the `@After` annotation like this:
```
@After
public void teardown() {
    inputStream.close();
}
```

aws-android-sdk-s3/src/test/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadInputStreamTest.java
```diff
@@ -0,0 +1,221 @@
...
+    @Test
+    public void testUpload() throws IOException {
+        String key = getRandomString();
+        transferUtility.upload(key, inputStream, UploadOptions.builder().build());
+        verify(transferUtility, times(1)).upload(eq(DEFAULT_BUCKET), eq(key), any(File.class),
+                any(ObjectMetadata.class), isNull(CannedAccessControlList.class), isNull(TransferListener.class));
+    }
+
+    @Test
+    public void testUploadWithBucket() throws IOException {
```
richardmcclellan, line 84: I rarely say there are too many tests, but in this case, I think there might be :).  Lots of testing is good, but we also want to keep in mind the maintainability of our test code.   For example, if we added a 5th parameter to the `Builder` in the future, we'd have to add 5 more tests if we want to follow the pattern that you've started here (and a 6th parameter would require 6 more tests, etc), but we really don't need that many to validate that our code works as expected.  As written, there is a lot of duplicate coverage of the same code paths.    Instead of testing every possible permutation, I think you could have three tests:

1. Test that `upload(String key, String inputStream)` calls the `File` based upload method with all of the expected defaults.  (It doesn't look like you have a test for this method at all right now actually)
2. Test that `upload(String key, String inputStream, UploadOptions options)` with a default Builder calls the `File` based upload method with all of the expected defaults.
3. Test that `upload(String key, String inputStream, UploadOptions options)` with all parameters specified calls the `File` based upload method with the same parameters that were provided as input.

aws-android-sdk-s3/src/test/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadInputStreamTest.java
```diff
@@ -0,0 +1,193 @@
...
+        String key = getRandomString();
+        transferUtility.upload(key, inputStream);
+        verify(transferUtility, times(1)).upload(eq(DEFAULT_BUCKET), eq(key),
+                any(File.class), any(ObjectMetadata.class), isNull(CannedAccessControlList.class),
+                isNull(TransferListener.class));
+    }
+
+    // Testing the upload(String key, InputStream inputStream, UploadOptions options) method calls
+    // upload(String bucket, String key, File file, ObjectMetadata metadata,
+    // CannedAccessControlList cannedAcl, TransferListener listener) in fact.
```
richardmcclellan, line 91:
```suggestion
```

aws-android-sdk-s3/src/test/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadInputStreamTest.java
```diff
@@ -0,0 +1,193 @@
...
+                .defaultBucket(DEFAULT_BUCKET)
+                .s3Client(s3)
+                .context(context)
+                .build());
+        inputStream = new ByteArrayInputStream("test data".getBytes());
+        metadata = new ObjectMetadata();
+        cannedAcl = CannedAccessControlList.Private;
+        listener = new UploadListener();
+    }
+
```
richardmcclellan, line 79:
```suggestion

/**
     * Test that upload(String key, String inputStream) calls the File based upload method with all of the expected defaults.
     * @throws IOException
     */
```

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferStatusUpdater.java
```diff
@@ -74,6 +75,11 @@
      */
     private static TransferStatusUpdater transferStatusUpdater;
 
+    /**
+     * Prefix of generated temporary file.
```
richardmcclellan, line 79:
```suggestion
     * Prefix for temporary File created when client uploads an `InputStream`.
```

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -762,6 +798,31 @@ private int createMultipartUploadRecords(String bucket, String key, File file, O
...
+            while ((bytesRead = inputStream.read(buffer)) != -1) {
+                outStream.write(buffer, 0, bytesRead);
+                outStream.flush();
+            }
+        } catch (IOException ioException) {
+            file.delete();
+            throw new IOException("Error writing the inputStream into a file.", ioException);
+        } finally {
+            outStream.close();
+        }
```
richardmcclellan, line 822: I think you can simplify this by just moving `file.delete()` to the `finally` block and getting rid of the `catch` block.  This ensures that the File will be deleted, and an `IOException` will be thrown up to the the client.

```suggestion
        } finally {
            file.delete();
            outStream.close();
        }
```

aws-android-sdk-s3/src/test/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadInputStreamTest.java
```diff
@@ -0,0 +1,193 @@
...
+            }
+        });
+
+        thread.start();
+
+        if (!deleteFileLatch.await(2, TimeUnit.SECONDS)) {
+            shadowOf(getMainLooper()).idle();
+            fail("File was not deleted after upload.");
+        }
+    }
```
richardmcclellan, line 165: While this test does pass now, expecting an `AssertionError` is essentially saying that the thing this test is supposed to be verifying, _doesn't_ work, so I don't think we want to do that. 

I looked into this a bit more and it looks like the upload was not getting started because there are some checks that verify the device has a network connection.  Since this is a unit test, and doesn't run on a device, those checks fail.    The goal of this test is only to verify that the temp file is deleted after upload, so instead of trying to setup a bunch of mocks to get a real upload to happen, I'd suggest just calling manually setting the state to COMPLETED, after calling `upload`, and then verifying that the file doesn't exist any more, like this:

```suggestion
    @Test
    public void testDeleteTempFileAfterUpload() throws IOException {
        // Call upload, so that a temporary File will be created
        final String key = getRandomString();
        observer = transferUtility.upload(key, inputStream);

        // Verify the file exists
        File file = new File(observer.getAbsoluteFilePath());
        assertTrue(file.exists());

        // Set state to COMPLETED
        Context context = InstrumentationRegistry.getInstrumentation().getContext();
        TransferStatusUpdater transferStatusUpdater = TransferStatusUpdater.getInstance(context);
        transferStatusUpdater.updateState(observer.getId(), TransferState.COMPLETED);

        // Verify the file was deleted
        assertFalse(file.exists());
    }
```

richardmcclellan (commented):
Just a couple more comments, but once those are addressed, ship it!

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
changxu0306: I think we cannot excecute `file.delete()` in `finally` block because we only want to delete a file when there is an exception occurs when writing into it. But if there is no exception, we want to keep that file. And delete it later when the transfer is completed.

richardmcclellan: Ahh!  Of course, that's right.  Ignore this comment :)

richardmcclellan:
From the unittest build, it looks like the 3 upload tests are failing with an `IllegalStateException`? Do you know why that is?

changxu0306:
> From the unittest build, it looks like the 3 upload tests are failing with an `IllegalStateException`? Do you know why that is?
It works fine on my machine 😱

aws-android-sdk-s3/src/test/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadInputStreamTest.java
```diff
@@ -112,6 +112,7 @@ public void testUploadWithAllParametersSet() throws IOException {
                                                               .build());
         verify(transferUtility, times(1)).upload(eq(bucket), eq(key), any(File.class),
                 eq(metadata), eq(cannedAcl), eq(listener));
+        transferUtility.getDbUtil().closeDB();
```
raphkim, line 115: `tearDown()` method is called after each test case, so this will be called later anyways. So let's just remove this line.

richardmcclellan (approved):
Looks good!   I know those unit tests were not easy, given the complexity of the existing architecture, - great job getting them working!

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferStatusUpdater.java
```diff
@@ -145,6 +151,15 @@ synchronized void removeTransfer(int id) {
...
     synchronized void removeTransferRecordFromDB(final int id) {
+        // Remove temporary file
+        TransferRecord transferRecord = dbUtil.getTransferById(id);
+        if (transferRecord != null) {
+            String path = transferRecord.file;
+            String fileName = new File(path).getName();
+            if (fileName.startsWith(TEMP_FILE_PREFIX)) {
+                new File(path).delete();
+            }
+        }
```
raphkim, line 162: This part worries me, because if a customer names their file `aws-s3-my-file`, then `TransferUtility` will delete it after upload even if it isn't temporary. Accidentally deleting customer data is probably a bigger issue than leaving a temp file sitting in their memory in my opinion. Is there a better way to mark a file as temporary?

raphkim (approved):
Looks good except one bit that worries me. Let's get this PR out of the way first, and then address it separately.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferStatusUpdater.java
changxu0306: I think maybe we could do the check with both the prefix and the suffix which is ".tmp". It's less possible for customers to upload a file with this kind of suffix. Or, make the prefix something like "aws-s3-generated-for-inputstream"?

richardmcclellan: Appending a constant UUID to the prefix would probably be a reasonable solution (e.g. `aws-s3-d861b25a-1edf-11eb-adc1-0242ac120002`)

I think it's actually quite reasonable that a customer could upload a file with the suffix `.tmp`, because that's the default suffix if none is provided when calling `File.createTempFile`, so I wouldn't rely on that.

changxu0306: When do checking before deleting a temp file, we don't know the UUID exactly. But if a random UUID indeed follows the format like `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`, we can adopt this method by checking whether the temp file fits into the format.

richardmcclellan: To clarify, I'm suggesting that you simply replace:
```
static final String TEMP_FILE_PREFIX = "aws-s3";
```
with 

```
static final String TEMP_FILE_PREFIX = "aws-s3-d861b25a-1edf-11eb-adc1-0242ac120002";
```

The chance that a customer would also prefix a file with `aws-s3` is very low -- but the chance they would prefix a File with `aws-s3-d861b25a-1edf-11eb-adc1-0242ac120002` is virtually none.

changxu0306: Got it. I thought each time we would generate a new UUID. Thanks! This totally make sense.

PR #2192, opened 2020-10-21
fix: remove outdated logging notes

jamesonwilliams:
These instructions are incorrect and misleading and need to be removed.


TrekSoft (approved):
Is there a reference to this document from another file that needs to also be deleted?

Cat16u3

PR #2198, opened 2020-10-28
support request host prefix in core sdk

raphkim:
Currently, the Android SDK is incapable of modifying the endpoint on a per-request basis. This PR allows request to remember a specific prefix to prepend to the request endpoint before execution without altering the client's endpoint itself.


aws-android-sdk-core/src/main/java/com/amazonaws/util/URIBuilder.java
```diff
@@ -0,0 +1,155 @@
+package com.amazonaws.util;
```
TrekSoft, line 1: I assume this is copied over from another module?

TrekSoft (approved):
Cat18u3

jamesonwilliams (approved):
In the `DefaultRequest`, can you kill the new Javadocs, and just rely on them cascading from `Request`? Possibly using [`{@inheritDoc}`](https://docs.oracle.com/javase/6/docs/technotes/tools/solaris/javadoc.html#@inheritDoc).

The documentation is a little abstract right now. I'm wondering: can you add an _example_ of what a host prefix is? Like, `s3.eu-west-1.amazonaws.com` vs... `famous-potatoss3.eu-west-1.amazonaws.com`? Maybe you can include a rationale for why a customer might want to use it -- what kinds of use cases it enabled for them?

aws-android-sdk-core/src/main/java/com/amazonaws/util/URIBuilder.java
jamesonwilliams: If that's the case, we should refactor that other module to use this new core capability, and then remove the duplicate.

raphkim: This structure is actually something I wrote up for this because I found it inconvenient to work with `URI` class. Is there an issue with the builder?

aws-android-sdk-core/src/main/java/com/amazonaws/util/URIBuilder.java
```diff
@@ -0,0 +1,155 @@
...
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+/**
+ * Utility class to help build a {@link URI} instance from scratch
+ * or from an existing URI. This is useful for replacing components
+ * inside an existing URI, which is immutable.
+ */
+public class URIBuilder {
```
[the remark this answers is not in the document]
raphkim: oh didn't even realize apache already has the exact same utility class!

TrekSoft (approved):
Cat19u4

PR #2214, opened 2020-11-04
updated prefix of temp file

changxu0306:


aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferStatusUpdater.java
```diff
@@ -78,7 +78,7 @@
     /**
      * Prefix for temporary File created when client uploads an `InputStream`.
      */
```
richardmcclellan, line 80:
```suggestion
    /**
     * Prefix for temporary File created when client uploads an InputStream.   When the 
     * upload completes, the File is deleted, if it was created by the library, but not if it was
     * provided by the client.   To ensure we don't delete any File objects provided by the client,
     * a random UUID is included in the prefix.
     */
```

richardmcclellan (dismissed):
Minor suggestion to add some more documentation, but otherwise :shipit: !

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferStatusUpdater.java
changxu0306: Agree that it's kinda important to add javadoc to it 😄

raphkim (approved):

richardmcclellan (approved):

Issue #2218, opened 2020-11-09
Typo: getSessionCredentitalsExpiration


mojodna:
[`CognitoCredentialsProvider.getSessionCredentitalsExpiration()`](https://github.com/aws-amplify/aws-sdk-android/blob/8018b729f6508944507eab9a097a50615c4576ba/aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCredentialsProvider.java#L443) is misspelled (should be `CognitoCredentialsProvider.getSessionCredentialsExpiration()`). Perhaps a fix for the next major version (and/or the addition of an alias in the next minor version)?

changxu0306:
Thank you for pointing out this! We will fix it asap.

changxu0306:
We have already fix this issue. It will be released in next version. I'll close this issue for now.

PR #2219, opened 2020-11-10
Fix misspelled method getSessionCredentialsExpiration()

changxu0306:
*Issue #, if available:*
Fix issue [#2218](https://github.com/aws-amplify/aws-sdk-android/issues/2218)


raphkim (changes requested):
Please make the new method contain the implementation rather than the other way around.

richardmcclellan:
There are 3 usages of the misspelled version in this library itself, so can you update those references to use the new one?

aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCredentialsProvider.java
```diff
@@ -440,6 +440,16 @@ public void setSessionCredentialsExpiration(Date expiration) {
...
+    public Date getSessionCredentialsExpiration() {
+        credentialsLock.readLock().lock();
+        try {
+            return sessionCredentialsExpiration;
+        } finally {
+            credentialsLock.readLock().unlock();
+        }
+    }
+
+    @Deprecated
```
richardmcclellan, line 452:
```suggestion
    @Deprecated Use {@link #getSessionCredentialsExpiration()} instead.
```

aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCredentialsProvider.java
```diff
@@ -440,6 +440,16 @@ public void setSessionCredentialsExpiration(Date expiration) {
...
     public Date getSessionCredentitalsExpiration() {
```
richardmcclellan, line 453: This method should call the new method, to avoid having the implementation logic duplicated in two places.

jamesonwilliams:
```java
public void newMethodWithGoodName() {
    // Contains the full implementation, that we will keep forever.
}

@Deprecated 
public void oldAndBadlyNamedMethod() {
    // It just forwards to the good method.
    methodWithCorrectName();
}
```

jamesonwilliams (dismissed):

aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCredentialsProvider.java
```diff
@@ -451,8 +451,8 @@ public Date getSessionCredentialsExpiration() {
 
     @Deprecated
     /** Use {@link #getSessionCredentialsExpiration()} instead. */
-    public Date getSessionCredentitalsExpiration() {
-        return getSessionCredentialsExpiration();
+    public void getSessionCredentitalsExpiration() {
```
richardmcclellan, line 454: Hmm, why change the return type here?  I think the way you had it before was correct.

aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCredentialsProvider.java
```diff
@@ -449,6 +449,12 @@ public Date getSessionCredentitalsExpiration() {
         }
     }
 
+    @Deprecated
+    /** Use {@link #getSessionCredentialsExpiration()} instead. */
```
raphkim, line 453: generally the annotation (`@Deprecated`) comes right before the method signature.

```suggestion
    /** 
     * @deprecated Use {@link #getSessionCredentialsExpiration()} instead. 
     */
    @Deprecated
```

aws-android-sdk-core/src/main/java/com/amazonaws/auth/CognitoCredentialsProvider.java
changxu0306: Oops! sorry I misunderstood Jameson's PR comment.

raphkim (approved):

raphkim (approved):

jamesonwilliams (approved):

PR #2220, opened 2020-11-11
remove modules that are no longer published

raphkim:
*Issue #, if available:*

*Description of changes:*
There is a CircleCI script that checks if all of the Android SDK packages are available on Maven, but this script hasn't been updated since some of the modules were removed. Updated the script to accurately reflect currently available packages.


richardmcclellan:
Hmm, are we supposed to use this script?  I haven't run it when doing the recent releases.   If so, maybe you could update the Quip doc describing how to release?

raphkim:
it is automatically run for any commit that is merged to main.

CircleciScripts/check_sdk_on_maven.py
```diff
@@ -47,7 +44,11 @@
     'aws-android-sdk-kinesis', 
     'aws-android-sdk-pinpoint', 
     'aws-android-sdk-machinelearning', 
-    'aws-android-sdk-sqs'
+    'aws-android-sdk-sqs',
+    'aws-android-sdk-connect',
+    'aws-android-sdk-connectparticipant',
+    'aws-android-sdk-textract',
+    'aws-android-sdk-kinesisvideo-signaling'
```
jamesonwilliams, line 51: This is fishy. If we're changing the order in some way, shouldn't we just do a lexicographical sort?

PR #2265, opened 2020-11-25
Fix NotAuthorizedException for new password required challenge

AliAsadi:
**Issue:**
When processing the `NewPasswordContinuation` (ChallengeName":"NEW_PASSWORD_REQUIRED")

The server returns NotAuthorizedException due to invalid `secretHash`.

It returns the following message: `Unable to verify secret hash for client XXXX`

**Fix Description:**
The client generates a secret hash combined of `email`, `clientId`, `clientSecret`
And the server generates a secret hash from the following `internal-username`, `clientId`, `clientSecret`.

client = `email`, `clientId`, `clientSecret`
server = `internal-username`, `clientId`, `clientSecret`

The server return NotAuthorizedException cause the secret hash in the client is different than the server. `client != server`.

To fix this we need to update the secret hash in the client using the server format. So the server can verify the secret hash. 

**Description of changes:**
Update the `secretHash` to use the server format, which is`internal-username`, `clientId`, `clientSecret`.

-----------------------


aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
```diff
@@ -3163,8 +3163,9 @@ public void run() {
             nextTask = deviceSrpAuthentication(clientMetadata, challenge, callback, runInBackground);
         } else if (CognitoServiceConstants.CHLG_TYPE_NEW_PASSWORD_REQUIRED.equals(challengeName)) {
             final NewPasswordContinuation newPasswordContinuation = new NewPasswordContinuation(
-                    cognitoUser, context, usernameInternal, clientId, secretHash, challenge,
-                    runInBackground, callback);
+                    cognitoUser, context, usernameInternal, clientId,
+                    CognitoSecretHash.getSecretHash(usernameInternal, clientId, clientSecret),
```
jamesonwilliams, line 3167: Does the `secretHash` itself need this update? Why only apply it here?

AliAsadi: Great question! Maybe but I'm not sure. To answer this, we need more information
from the server-side about the usage of the client secret-hash.

Is there any use for the client secret-hash after the AuthenticationContinuation? or we can override it?

I mean by client secret-hash is this.
client = `email`, `clientId`, `clientSecret`.

And I'm applying it only here because I am working on the reset password challenge
and faced this problem. But it could be relevant for other challenges as well worth a check.

AliAsadi: Great question! Maybe but I'm not sure. To answer this, we need more information
from the server-side about the usage of the client secret-hash.

Is there any use for the client secret-hash after the AuthenticationContinuation? or we can override it?

I mean by client secret-hash is this.
client = `email`, `clientId`, `clientSecret`.

And I'm applying it only here because I am working on the new password required challenge
and faced this problem. But it could be relevant for other challenges as well worth a check.

jamesonwilliams: I'll forward this to the Cognito team and see if anyone has guidance for us.

jamesonwilliams: Cognito team reviewed this internally and approved.

jamesonwilliams (approved):

PR #2277, opened 2020-12-08
chore: remove apache commons logging dependency

raphkim:
*Description of changes:*
Currently, Android Studio complains that:

> `commons-logging` defines classes that conflict with classes now provided by Android. Solutions include finding newer versions or alternative libraries that don't have the same problem (for example, for `httpclient` use `HttpUrlConnection` or `okhttp` instead), or repackaging the library using something like `jarjar`.

This PR removes gradle dependency on `'commons-logging:commons-logging:1.2'` and deprecates `ApacheCommonsLogging` class in favor of always using `AndroidLog` class.

Some unit tests had to be modified to run with `RobolectricTestRunner` due to this change, because JVM cannot run Android code (`android.util.Log`) by itself.

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/logging/ApacheCommonsLogging.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/logging/ApacheCommonsLogging.java
@@ -15,14 +15,14 @@
 
 package com.amazonaws.logging;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
 import com.amazonaws.logging.LogFactory.Level;
 
 /**
  * This is a wrapper over the Apache Commons
  * Logging framework.
+ * @deprecated Use {@link AndroidLog} instead.
  */
+@Deprecated
 public class ApacheCommonsLogging implements com.amazonaws.logging.Log {
 
     private Class logClass;
--- a/Proguard.html
+++ b/Proguard.html
@@ -90,7 +90,6 @@ the following lines to the end of your <i>proguard-project.txt</i> file:
 -keep class com.amazonaws.services.**.*Handler
 # The following are referenced but aren't required to run
 -dontwarn com.fasterxml.jackson.**
--dontwarn org.apache.commons.logging.**
 # Android 6.0 release removes support for the Apache HTTP client
 -dontwarn org.apache.http.**
 # The SDK has several references of Apache HTTP client
--- a/aws-android-sdk-core/build.gradle
+++ b/aws-android-sdk-core/build.gradle
@@ -14,7 +14,6 @@ android {
 
 dependencies {
     api 'com.google.code.gson:gson:2.8.6'
-    compileOnly 'commons-logging:commons-logging:1.2'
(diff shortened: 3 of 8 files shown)


aws-android-sdk-core/src/main/java/com/amazonaws/logging/ApacheCommonsLogging.java
```diff
@@ -15,14 +15,14 @@
...
-import org.apache.commons.logging.LogFactory;
 import com.amazonaws.logging.LogFactory.Level;
 
 /**
  * This is a wrapper over the Apache Commons
  * Logging framework.
+ * @deprecated Use {@link AndroidLog} instead.
  */
+@Deprecated
 public class ApacheCommonsLogging implements com.amazonaws.logging.Log {
```
jamesonwilliams, line 26: Can we delete the file?

jamesonwilliams (approved):
Woohoo nice! Do you want to add the console log first, to get rid of the need for these Robolectric additions?

aws-android-sdk-core/src/main/java/com/amazonaws/logging/ApacheCommonsLogging.java
raphkim: Yea, I doubt any customer was using this explicitly... but still I feel iffy about downright deleting a public class

jamesonwilliams (changes requested):
We chatted offline and there are a few more usages around the code base to address:

```console
find /Users/jhwill/aws-sdk-android \( -name "*.java" -o -name "*.gradle" \) -exec grep -l '.*commons.*log.*' {}  + 
/Users/jhwill/aws-sdk-android/aws-android-sdk-s3-test/build.gradle
/Users/jhwill/aws-sdk-android/aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/internal/crypto/CryptoTestUtils.java
/Users/jhwill/aws-sdk-android/aws-android-sdk-core/build.gradle
/Users/jhwill/aws-sdk-android/aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
/Users/jhwill/aws-sdk-android/aws-android-sdk-core/src/main/java/com/amazonaws/logging/ApacheCommonsLogging.java
/Users/jhwill/aws-sdk-android/aws-android-sdk-s3/build.gradle
```

aws-android-sdk-core/src/main/java/com/amazonaws/logging/ApacheCommonsLogging.java
```diff
@@ -15,14 +15,14 @@
 
 package com.amazonaws.logging;
 
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
```
jamesonwilliams, line 19: Oh man, I see. This was shadowing the package local copy? And this document still compiles, because now `LogFactory` refers to the package local class? Creepy stuff, sigh.

jamesonwilliams (approved):
Beautiful stuff.

PR #2278, opened 2020-12-08
chore: refactor LogFactory

raphkim:
*Description of changes:*
Due to an extensive use of `LogFactory` (which is implemented using `android.util.Log`) throughout the project, the following runtime exception tends to pop up from place to place while running unit tests: 
```
Caused by: java.lang.RuntimeException: Method w in android.util.Log not mocked. See http://g.co/androidstudio/not-mocked for details.
```
This can be bypassed generally by using `robolectric` package, but this also slows down the unit test execution significantly.

This PR addresses the above problem by detecting if unit test is unable to mock static methods from `android.util.Log` and defaulting to another logger implementation that outputs to console instead.


aws-android-sdk-core/src/main/java/com/amazonaws/logging/ConsoleLog.java
```diff
@@ -0,0 +1,156 @@
...
+ * permissions and limitations under the License.
+ */
+
+package com.amazonaws.logging;
+
+/**
+ * Simple log to print to console. It conforms to the
+ * {@code com.amazonaws.logging.Log} interface.
+ */
+public class ConsoleLog implements com.amazonaws.logging.Log {
```
jamesonwilliams, line 22:
```suggestion
public final class ConsoleLog implements com.amazonaws.logging.Log {
```

aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
```diff
@@ -53,20 +54,18 @@ public static synchronized Log getLog(String logTag) {
...
-            }
-            if (log == null) {
-                log = new AndroidLog(logTag);
-                logMap.put(logTag, log);
-            }
+        if (log != null) {
+            return log;
+        }
+
+        if (checkApacheCommonsLoggingExists()) {
```
jamesonwilliams, line 61: I would just get of any semblance of support for Apache.

jamesonwilliams (commented):
Richard did some work to toggle between Android & console logging implementations for the Amplify Android library. That has been working pretty well and similarly addresses the goal of not needing Robolectric everywhere: https://github.com/aws-amplify/amplify-android/pull/744/files

aws-android-sdk-core/src/main/java/com/amazonaws/logging/LogFactory.java
raphkim: I will do this in a separate PR in case there is even the smallest chance of needing to revert the commit :)
Will be followed up by #2277

jamesonwilliams (approved):
I guess there's no easy way to know what test _still_ need Robolectric. But I suspect that you can remove many `@RunWith` statements, now, and speedup the test runs. That'll be sort of tedious, to strip them all off, and individually add them back in when you get non-logging-related test double issues. But, this would be a good time to do the exercise, if we're going to.

I understand this PR to be logically different than removing Apache. But, that will be a good _next_ improvement to this logging code, once the tests aren't falling back to the (unavailable) Android logger. 👍

jamesonwilliams (approved):

Issue #2284, opened 2020-12-11
Remove Cognito (Sync), Mobile Analytics, and S3 TransferManager


jamesonwilliams:
The [2.19.0 release of the AWS Mobile SDK for iOS](https://github.com/aws-amplify/aws-sdk-ios/blob/49d1ec7b2e4df212469142ca66ad693698a58285/CHANGELOG.md#2200) removed the following modules:
1. AWS Cognito (Sync)
2. AWS Mobile Analytics
3. AWS S3 Transfer Manager

The AWS Mobile SDK for Android should also remove these modules. Our equivalent versions of them are here:
1. [`aws-android-sdk-cognito`](https://github.com/aws-amplify/aws-sdk-android/tree/main/aws-android-sdk-cognito)
2. [`aws-android-sdk-mobileanalytics`](https://github.com/aws-amplify/aws-sdk-android/tree/main/aws-android-sdk-mobileanalytics)
3. [`transfermanager` in `aws-android-sdk-s3`](https://github.com/aws-amplify/aws-sdk-android/blob/main/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transfermanager)

PR #2285, opened 2020-12-11
chore: remove transfer manager

raphkim:
*Issue #, if available:*
#2284 

*Description of changes:*

**Breaking changes**
* Remove `transfermanager`, and consequently, it's usage from `DynamoDB` module.
  * Remove `getTransferManager()` method from `S3ClientCache` 
  * `S3ClientCache` no longer caches `TransferManager`; only `AmazonS3Client` is cached. This class provided the former for customer's convenience, but it did not actually use the transfer manager it cached anywhere in the codebase. Its main purpose was to cache the S3 client.

Affected test files:
* CopyPartIntegrationTest
  * Removed test that references `TransferManager`, which was used to upload a large file to copy from
  * The tests currently pass regardless of error status, so fixed the test to _actually_ assert that they pass
* MultiPartUploadsIntegrationTest
  * Removed reference to `TransferManager`
* AmazonS3ClientIntegrationTest.java
  * Made bucket creation and deletion steps required (currently ignores if either fails)


CircleciScripts/ReleaseConfiguration.json
```diff
@@ -14,7 +14,6 @@
 			"aws-android-sdk-auth-userpools",
 			"aws-android-sdk-autoscaling",
 			"aws-android-sdk-cloudwatch",
-			"aws-android-sdk-cognito",
 		    // "aws-android-sdk-cognitoauth",
```
jamesonwilliams, line 17: Hm, this doesn't look good. `//` isn't valid in the middle of a JSON document.

aws-android-sdk-ddb-mapper/src/main/java/com/amazonaws/mobileconnectors/dynamodbv2/dynamodbmapper/S3ClientCache.java
```diff
@@ -109,39 +74,9 @@ public AmazonS3Client getClient(Region s3region) {
...
-     * returned could be shut down if a new underlying {@link AmazonS3Client} is
-     * provided with {@link #useClient(AmazonS3Client)}.
-     *
-     * @param region The region the returned {@link TransferManager} will be
-     *            configured to use.
-     * @return A transfer manager for the given region from the cache, or one
-     *         instantiated automatically from any existing
-     *         {@link AmazonS3Client},
-     */
-    public TransferManager getTransferManager(Region region) {
```
jamesonwilliams, line 137: TLDR, it wasn't actually being used by the mapper? Or why is this safe?

jamesonwilliams (changes requested):
Looks good, I think. The straight code removals are clear. But, can you add more description to the code _changes_ in the PR description? Also, there's a weird thing going on with a `//` in a JSON doc. JSON don't support these types of comments.

CircleciScripts/ReleaseConfiguration.json
raphkim: hmm, it was always there.. that _is_ strange though. You are right that JSON doc doesn't support this on hindsight, but I didn't see anything wrong with it until you just mentioned haha

aws-android-sdk-ddb-mapper/src/main/java/com/amazonaws/mobileconnectors/dynamodbv2/dynamodbmapper/S3ClientCache.java
raphkim: It was not being used outside of integration tests. Since this method directly returns a deprecated item, I think it is within scope to remove this too.

The main purpose of this class was to map S3 clients by region. It also associated each of those clients to a transfer manager instance, but this was just to make it easier for customers to use those S3 clients. Nowhere in the codebase was the transfer manger being explicitly used though.

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
```diff
@@ -451,11 +451,14 @@ public TransferObserver upload(String bucket, String key, File file) {
...
      *
+     *
+     * @param largeBucketName
      * @param key  The key in the specified bucket by which to store the new object.
      * @param file The file to upload.
+     * @param transferListener
      * @return A TransferObserver used to track upload progress and state
      */
-    public TransferObserver upload(String key, File file) {
+    public TransferObserver upload(String largeBucketName, String key, File file, TransferListener transferListener) {
```
jamesonwilliams, line 461: What's the deal here? Is this a breaking change?

aws-android-sdk-s3-test/src/androidTest/java/com/amazonaws/services/s3/CopyPartIntegrationTest.java
```diff
@@ -52,22 +48,22 @@
 public class CopyPartIntegrationTest extends S3IntegrationTestBase {
```
jamesonwilliams, line 48: Can you include a description of the changes to this file in the PR description?

jamesonwilliams (commented):
What are the changes in `CopyPartIntegrationTest` and `TransferUtility`?

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtility.java
raphkim: oof where the heck did this come from? let me check

raphkim: fixed.

jamesonwilliams:
Any updates on this work?

raphkim:
Not yet. I'll take another look today.

jamesonwilliams:
There are a number of different things going on in this PR, now.

Let's take a more bite-sized approach.

Here's a PR to remove the Cognito Sync client: https://github.com/aws-amplify/aws-sdk-android/pull/2320

Here's a PR to remove the Mobile Analytics client: https://github.com/aws-amplify/aws-sdk-android/pull/2321

Let's rebase this current PR on top of those, and make it just about `TransferManager` changes.

richardmcclellan (approved):
🚢

jamesonwilliams (approved):

kenyee:
@raphkim Trying to migrate to TransferUtility since TransferManager was ripped out of the latest SDK.

What's the replacement for getting the e-tag of the uploaded file?
```
            val tx = TransferManager(s3Client)
            val upload = tx.upload(awsInfo.bucketName, s3FileKeyName, file)
            val result = upload.waitForUploadResult()
            Timber.d("Upload complete.  E-tag: ${result.eTag}")
```

There's an old comment at the bottom of this blog asking about that too which is why I never migrated to TransferUtility.

jamesonwilliams:
@kenyee I don't think `TransferUtility` exposes the ETag, but you can access it through the S3 client directly:

```kotlin
fun eTag(bucketName: String, key: String) {
    return AmazonS3Client(AWSMobileClient.instance)
        .getObject(GetObjectRequest(bucketName, key))
        .objectMetadata
        .eTag
```

kenyee:
@jamesonwilliams 
Sorry for the late reply...finally found some time to try this and our use-case doesn't work with that technique...filed an issue on it FYI:
https://github.com/aws-amplify/aws-sdk-android/issues/2348

TLDR: our upload bucket is write only w/ a temp key for security reasons...can't read the etag after the upload because there's no read access to it...

jamesonwilliams:
> @jamesonwilliams 
> 
> Sorry for the late reply...finally found some time to try this and our use-case doesn't work with that technique...filed an issue on it FYI:
> 
> https://github.com/aws-amplify/aws-sdk-android/issues/2348
> 
> 
> 
> TLDR: our upload bucket is write only w/ a temp key for security reasons...can't read the etag after the upload because there's no read access to it...

Hey @kenyee, if you have unusual requirements like write-only, you may want to migrate to raw use of the AmazonS3Client, and/or create a purpose built API with API Gateway and the code your S3 logic behind it.

kenyee:
looks like I can use S3Client.putObject() and get the etag out of the objectResult.  Thx.
I'd lose the progress listener (doesn't seem to be a hook for that in the S3 client except for the monitoringListener that doesn't seem documented very well) but I don't need that anyways 😃

PR #2325, opened 2021-01-09
fix(aws-android-sdk-iot): Improve AWS Region code extraction

jpignata:
*Issue #, if available:* n/a

*Description of changes:* Previously this helper was looking for an AWS Region code within a static component of the endpoint. AWS IoT Device Advisor introduces a new format for test endpoints. Rather than adding to this logic, this change tokenizes the endpoint and looks for a matching AWS Region code.


aws-android-sdk-iot/src/main/java/com/amazonaws/mobileconnectors/iot/AwsIotEndpointUtility.java
```diff
@@ -7,50 +7,29 @@
  * Helper class for parsing AWS IoT endpoints.
  */
 public final class AwsIotEndpointUtility {
```
jpignata, line 9: It's likely this helper class doesn't need to exist. It's also equally likely the client is doing this extraction unnecessarily.

jpignata: I moved the method from this class into the client itself. The extraction is necessary for IAM authentication, apparently.

rjuliano (approved):
🚢  it!!!

PR #2363, opened 2021-02-20
fix(cognitoauth): Prevent NPE on HostedUI flow

rjuliano:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/1090

*Description of changes:*
I haven't been able to reproduce [this issue](https://github.com/aws-amplify/amplify-android/issues/1090), which was actually reported in amplify-android. This is an attempt at resolving that, but without being able to reproduce it, it's hard to know with any certainty if it will.

### DIFF
--- a/aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/activities/CustomTabsManagerActivity.java
+++ b/aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/activities/CustomTabsManagerActivity.java
@@ -99,6 +99,14 @@ public final class CustomTabsManagerActivity extends Activity {
         outState.putParcelable(CUSTOM_TABS_INTENT_KEY, customTabsIntent);
     }
 
+    @Override
+    protected void onRestoreInstanceState(Bundle savedInstanceState) {
+        super.onRestoreInstanceState(savedInstanceState);
+        if (savedInstanceState != null) {
+            extractState(savedInstanceState);
+        }
+    }
+
     private void handleAuthorizationComplete() {
         setResult(RESULT_OK, getIntent());
     }


richardmcclellan:
I did some testing with this change, but I'm not sure how it will help.  The following sequence occurs on sign in:

```
CustomTabsManagerActivity: onCreate, bundle: null
CustomTabsManagerActivity: extractState customTabsLaunched:  false
CustomTabsManagerActivity: onResume firstRun! starting activity
CustomTabsManagerActivity: onSaveInstanceState
auth-hostedui:SignInActivity: SignInActivity onCreate
CustomTabsManagerActivity: onCreate, bundle: NOT null
CustomTabsManagerActivity: extractState customTabsLaunched:  true
CustomTabsManagerActivity: onRestoreInstanceState
CustomTabsManagerActivity: extractState customTabsLaunched:  true
CustomTabsManagerActivity: onNewIntent
CustomTabsManagerActivity: onResume, not first run, intent has data
```

`onRestoreInstanceState is called immediately after `onCreate`, with the state `Bundle` passed to each, so now we are just unpacking the same bundle twice, which is redundant.

The following sequence occurs on sign out:

```
auth-hostedui:MainActivity: signOut
IdentityManager: Signing out...
AWSMobileClient: Inspecting user state details
auth-hostedui:MainActivity: sign out succeeded
auth-hostedui:SignInActivity: SignInActivity onCreate
CustomTabsManagerActivity: onCreate, bundle: null
AuthClient: CustomTabsManagerActivity was created with a null state. This should never happen. Please file an issue with us on GitHub.
```

Implementing `onRestoreSavedInstance` state appears to have no impact on sign out, since it is not called during this flow. 

I'm trying to figure out if the AuthClient error being logged is an actual issue.  We should fix that if it is, or not log that message if it's not a problem.

PR #2368, opened 2021-02-23
Issue1003: registerConfigSignInProviders in _initialize resolves signInProviders missing for DropInUI

BillBunting:
To resolve, AWSMobileClient.getInstance().showSignIn() buttons do not work after sign-out or invalid token, issue #1003. The problem was created by changes made in 2.13.4. 

*Issue #* 1003

*Description of changes:* By moving the call to `registerConfigSignInProviders()` up to _initialize in AWSMobleClient.java resoles the problem when MobileClient "Failed to federate the tokens. Token expired:". MobileClient return SIGN_IN state then immediately returns SIGNED_OUT_FEDERATED_TOKENS_INVALID to the user state listener. 

`registerConfigSignInProviders()` sets the sign in providers, but unless set prior to `_showSignInDropInUI(callingActivity`, the providers are not found when needed by the Drop in UI, in some use cases such as when federated tokens are invalid. 

### DIFF
--- a/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
+++ b/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
@@ -457,6 +457,7 @@ public final class AWSMobileClient implements AWSCredentialsProvider {
     protected Runnable _initialize(final Context context, final AWSConfiguration awsConfiguration, final Callback<UserStateDetails> callback) {
         return new Runnable() {
             public void run() {
+
                 synchronized (initLockObject) {
                     if (AWSMobileClient.this.awsConfiguration != null) {
                         callback.onResult(getUserStateDetails(true));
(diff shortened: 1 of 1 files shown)


aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -483,11 +484,14 @@ public void run() {
                     mContext = context.getApplicationContext();
                     mStore = new AWSMobileClientStore(AWSMobileClient.this);
 
+                    AWSMobileClient.this.awsConfiguration = awsConfiguration;
```
raphkim, line 487: This is intentionally done last in `_initialize()` to validate that the provided configuration is correct. It seems that you are doing this ahead of time so that `registerConfigSignInProviders()` can take effect. I think the better approach would be to let `registerConfigSignInProviders()` method take in a configuration object as a parameter and pass it in.

raphkim:
Thank you for your contribution @BillBunting! 
This is great as always and looks like it should be good to go after addressing a relatively minor item.

raphkim:
Hi @BillBunting , I addressed my own comment in #2386. I will close this thread and track the new one.

BillBunting:
@raphkim Thank you for working on this!

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
BillBunting: @raphkim I had been using my local branch version and recently switched back to using the latest implementation via gradle dependency on 2.23.0. The problem returned. Without setting AWSMobileClient.this.awsConfiguration = awsConfiguration at the line shown, earlier than is done in the latest, there are use cases where this.awsConfiguration is null, causing the UX to fail to initialize properly and register sign in providers. I will update my local copy and test.

raphkim: By design, `AWSMobileClient#awsConfiguration` should be `null` if `AWSMobileClient#initialize()` did not trigger a success callback yet. This is to prevent any accidental usage of an invalid configuration that previously failed to initialize mobile client in other parts of the code. If there is any need for configuration object during the initialization process, then it should use the configuration instance that was passed in as a parameter rather than the instance variable of mobile client instance.

Issue #2349, opened 2021-01-31
Switching Auth flow to USER_PASSWORD_AUTH is not working


kapi-ai:
**Describe the bug**
I am using the Amplify library in an App written in Kotlin. While experimenting with the "authenticationFlowType" setting in my amplifyconfiguration.json file, I noticed that it is impossible to switch to USER_PASSWORD_AUTH, even, if you set it in the amplifyconfiguration.

The problem here lies in  the Sign-In functionality of the AWSMobileClient.java of the aws-android-sdk-mobile-client. The function _sign_in internally overrides the function getAuthenticationDetails() in line 1207 (in tag 2.22.1). 
In this override, new AuthenticationDetails are created depending on the awsConfiguration. New AuthenticationDetails are created and the default constructor always uses "CognitoServiceConstants.CHLG_TYPE_USER_PASSWORD_VERIFIER" as authenticationType so that is what we end up with in the AuthenticationDetails.
Now, when we want to initiate the user authentication in _initiateUserAuthentication() in line 1178 of CognitoUser.java we check using some if/elseif blocks the authenticationType. Since it is by default always  CHLG_TYPE_USER_PASSWORD_VERIFIER we can never go to the function startWithUserPasswordAuth() but always go to startWithUserSrpAuth().

The root of the problem seems to be hardcoding the value for authenticationType when constructing new AuthenticationDetails and then not checking them after creation. This should be checked or changed. 

**To Reproduce**
A code sample or steps:
amplifyconfig.json
```
{
  "auth": {
    "plugins": {
      "awsCognitoAuthPlugin": {
        "IdentityManager": {
          "Default": {}
        },
        "CredentialsProvider": {
          "CognitoIdentity": {
            "Default": {
              "PoolId": "XXX",
              "Region": "XXX"
            }
          }
        },
        "CognitoUserPool": {
          "Default": {
            "PoolId": "XXX",
            "AppClientId": "XXX",
            "Region": "XXX"
          }
        },
        "Auth": {
          "Default": {
            "authenticationFlowType": "USER_PASSWORD_AUTH"
          }
        }
      }
    }
  }
}
```
Login part in AuthActivity.kt:
```
     Amplify.Auth.signIn(
            username,
            password,
            { result ->
                runOnUiThread(Runnable {
                    Toast
                        .makeText(
                            this@AuthenticatorActivity,
                            "Logged in",
                            Toast.LENGTH_LONG
                        )
                        .show()
                })
            },
[...]

Issue #1778, opened 2020-05-22
How to initiate "USER_PASSWORD" sign in flow in android.


sergiotorresperez:
We are trying to migrate the cognito user pool in our system.
To do this, we do need to use "USER_PASSWORD" AuthenticationType, as explained [here](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-dg.pdf): 

> Your app sends the username and password to Amazon Cognito. If your app has a native sign-in
> UI and uses the Cognito Identity Provider SDK, your app must use the USER_PASSWORD_AUTH
> flow, in which the SDK sends the password to the server (your app must not use the
> default USER_SRP_AUTH flow since the SDK does not send the password to the server
> in the SRP authentication flow). The USER_PASSWORD_AUTH flow is enabled by setting
> AuthenticationDetails.authenticationType to "USER_PASSWORD".

The [Amplitude documentation](https://docs.aws.amazon.com/cognito/index.html) does not explain how to do this.

[This thread ](https://github.com/aws-amplify/aws-sdk-android/issues/1051) states we need to use 
`android-sdk-cognitoidentityprovider` SDK in order to specify the `AuthenticationType`.

1. - Could you please refer me to one example or code snipped on how to implement this use case using the aforementioned SDK? I can't understand how to do this with the javadoc documentation referred - The javadocs explain how to build one `AuthenticationDetails` object with the desired authentication type, but I don't know what to do with it. 
How could I effectively start the authentication flow with the manually constructed `AuthenticationDetails`?

2. - Our app is currently using the Amplitude SDK extensively for sign-in and authentication features. 
If we use `android-sdk-cognitoidentityprovider` to sign in with  "USER_PASSWORD", will the existing code be compatible with it? 
For example, will the app be able to get the id tokens using `AwsMobileClient.getTokens()`, will the automatic refreshment of expired tokens work, and will the "reset forgotten password" use case work? 

**Which AWS Services are you utilizing?**

com.amazonaws:aws-android-sdk-mobile-client:2.16.10
com.amazonaws:aws-android-sdk-s3:2.16.10
com.amazonaws:aws-android-sdk-cognitoauth:2.16.10

**Environment(please complete the following information):**
 - SDK Version: [e.g. 2.6.25]
2.16.10

**Device Information (please complete the following information):**
 - Device: Any android phone
 - Android Version: minSdkVersion is API 26
 - Specific to simulators: No.

OsmarICancino:
[...]

PR #2377, opened 2021-03-09
fix: honor authenticationFlowType from config

rjuliano:
*Issue #, if available: #1778 #2349

*Description of changes:
Currently, the value from `authenticationFlowType` does not get used properly inside mobile client. This PR attempts to close that gap by using the value specified in the config.

We have a couple issues open for this right now. There may be some in the amplify-android repo as well.

Some things to consider:
- This could be considered a change in behavior since the new code would start picking up the value from the config rather than defaulting to USER_SRP_AUTH
- Looking at the iOS code, I don't see any code to use that setting beyond determining whether it's custom auth. Need to do a little more digging on that.


aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -1206,20 +1206,33 @@ public void onSuccess(CognitoUserSession userSession, CognitoDevice newDevice) {
...
                                 try {
-                                    if (
-                                            awsConfiguration.optJsonObject(AUTH_KEY) != null &&
-                                            awsConfiguration.optJsonObject(AUTH_KEY).has("authenticationFlowType") &&
-                                            awsConfiguration.optJsonObject(AUTH_KEY).getString("authenticationFlowType").equals("CUSTOM_AUTH")
-                                    ) {
-                                        final HashMap<String, String> authParameters = new HashMap<String, String>();
+                                    // If there's a value in the config and it's CUSTOM_AUTH
+                                    if (authFlowTypeInConfig &&
+                                        awsConfiguration.optJsonObject(AUTH_KEY).getString("authenticationFlowType").equals("CUSTOM_AUTH")) {
```
raphkim, line 1220: `"CUSTOM_AUTH".equals(otherstuff)` to protect against `null`. What happens if value `null` is saved as value? would `getString()` just throw since it's not a string?

jamesonwilliams (approved):
🔥

Abhishek0786:
@rjuliano
your changes has been reverted in 2.22.5 version of aws-sdk-android.
can anybody help me how to solve this problem in latest version 2.22.6 of aws-sdk-android?

@raphkim Could you please explain why you have reverted those changes and is there any other workaround for this?

raphkim:
Hi @Abhishek0786, 

We found that there was an unintended behavior introduced by this commit, which required us to revert the changes until we could further investigate it. We are hoping to make some adjustments to the foundations laid out by this PR to fully address the issue in near future.

Abhishek0786:
Hi @raphkim,

Thanks for the information.
But this issue is still there and I am not able to migrate the cognito users in android but same steps is working fine in iOS. In android lambda function is not triggering when user try to login but when user do the same for forgot password then lambda is triggering and user is migrating.
 
Note: I am using the same awsconfiguration.json file and using the AWSMobileClient SDK in Android and iOS.

**For your reference:**

{
"IdentityManager": {
"Default": {}
},
"CognitoUserPool": {
"Default": {
"PoolId": "XX-XXXXXX_XXXXM",
"AppClientId":"XXXXXXXXXXXXXXXXXXXXXXXXXX",
"Region": "XX-XXXXXX" ,
"MigrationEnabled": "True"
}
},
"Auth": {
"Default": {
"authenticationFlowType": "USER_PASSWORD_AUTH"
}
}
}

Issue #2268, opened 2020-11-26
NullPointerException in AWSMobileClient.signOut


imetodieva:
I'm getting NullPointer from AWSMobileClient.signOut function 

**To Reproduce**
Call to AWSMobileClient.getInstance().signOut()

**Which AWS service(s) are affected?**
AWSMobileClient

**Expected behavior**
No crash happening.

**Environment Information :**
 - com.amazonaws:aws-android-sdk-mobile-client:2.18.0
 - Devices: [Huawei Mate 20 RS , Samsung Galaxy S10,Samsung Galaxy Note8,Huawei HUAWEI P smart+ 2019,Samsung Galaxy A51 etc.]
 - Android Version: [Android 9 Pie, Android 10, Android 11]
 - Specific to simulators: [No]

**Additional context**
Not happening on emulators.
App min version is 21.
I can only share part of the stacktrace :

Caused by: java.lang.NullPointerException: 
  at com.amazonaws.mobile.client.AWSMobileClient.signOut (AWSMobileClient.java:1467)

shikhar-91s:
Was this working fine previously? If this was working fine previously and you are getting this now all of a sudden, then it might be related to recent Aws Outage. For more info: https://www.theverge.com/2020/11/25/21719396/amazon-web-services-aws-outage-down-internet.

shikhar-91s:
As you have only shown part of the stack trace it seems to me its a Null Pointer Exception on reference returned by getInstance(). Just check whether AwsMobileClient.getInstance() (probably a singleton) is null or not using debugger. You might be lazily initialising it somewhere in code but you called sign out first without initialising it properly.

imetodieva:
I haven't seen this problem before. From the stacktrace I see that the problem lies within **mStore** field of the AWSMobileClient.java being null . If the singleton wasn't initialized other earlier calls to the library would be problematic but they aren't. And for the part with the outage - I don't think that is it because it's not making the global signout of all sessions, the code doesn't seem to make any other thing than to clear out local data. If it were something other than that **mStore** field wouldn't that mean that the **mFederatedLoginsMap.clear()** would also fail with nullpointer? This is assumption made by examining the method from the library that is invoked:

/**
     * Clears local tokens so that the client is in a signed-out state.
     */
    @AnyThread
    public void signOut() {
        mCognitoUserSession = null;
        if (userpool != null) {
            userpool.getCurrentUser().signOut();
            userpool.getUser().signOut();
        }
[...]

PR #2383, opened 2021-03-16
initialize key value store inside mobile client constructor

raphkim:
*Issue #, if available:*
#2268

*Description of changes:*
`mStore` variable is used throughout the `AWSMobileClient` codebase without null-check even though it is not initialized until after `AWSMobileClient#initialize()` is called. This PR adds a dummy store to initialize the variable with in order to prevent any potential NPE.  


aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -3789,7 +3790,49 @@ public SignInProviderConfig(final Class<? extends SignInProvider> signInProvider
     }
 }
 
-class AWSMobileClientStore {
+interface KeyValueStore {
```
richardmcclellan, line 3793: Could this be in a new file?  I know the previous pattern here seems to be to throw everything into this one massive [4060 line file](https://github.com/aws-amplify/aws-sdk-android/blob/ac38a5b18fb546c54251279e07b24d84c70f3ece/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java#L4060), but perhaps we could break the trend now? :)

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -3789,7 +3790,49 @@ public SignInProviderConfig(final Class<? extends SignInProvider> signInProvider
...
+
+    }
+
+    @Override
+    public void clear() {
+        throwNotInitializedException();
+    }
+
+    private void throwNotInitializedException() {
+        throw new RuntimeException("AWSMobileClient has not been initialized yet.");
```
richardmcclellan, line 3831: So, it looks like this PR doesn't fix the crash, it just replaces the NPE with this more specific crash, which could help us to identify the real root cause later, is that right?

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
raphkim: haha yea, we can do that

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
raphkim: yep. the customer will still come across an error, but we'll know if it was caused by faulty initialization or something else.

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -327,6 +327,7 @@ private AWSMobileClient() {
         federateWithCognitoIdentityLockObject = new Object();
         showSignInWaitLatch = new CountDownLatch(1);
         initLockObject = new Object();
+        mStore = new DummyStore();
```
jamesonwilliams, line 330: Seems like the root issue here is that `mStore` is being used before it is initialized. Can we either:
1. Initialize `mStore` with its real implementation, _earlier_?
2. Prevent the calls from happening to the `mStore`, _until_ it exists?

raphkim: 1. `AWSMobileClientStore` requires context, which is acquired during initialization, so the real implementation cannot exist until after initialization
2. This would have been ideal, but there are a lot of calls to `mStore` throughout the codebase, which makes it unrealistic to address them all

Strangely enough, the customer reporting the issue claimed that mobile client was successfully, so this PR will help us better identify the root cause.

jamesonwilliams (approved):

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClientStore.java
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+
+package com.amazonaws.mobile.client;
+
+import com.amazonaws.internal.keyvaluestore.AWSKeyValueStore;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.locks.ReadWriteLock;
```
jamesonwilliams, line 24: Are these imports still in the other file? Can you run optimize imports on the old file?

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/DummyStore.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package com.amazonaws.mobile.client;
+
+import java.util.Map;
+
+/**
+ * Default key value store to be replaced during mobile client initialization.
+ */
+final class DummyStore implements KeyValueStore {
```
jamesonwilliams, line 25: Hm, I'm not understanding.

Previously, we would throw an NPE if the store hadn't been initialized.

Now, we have a non-null store, and it throws a not-initialized exception.

Couldn't we just catch the NPE and wrap it?

raphkim: That means we would need to try-catch all 30 unguarded usage of `mStore` inside `AWSMobileClient`, which I find to be more troublesome than this approach.

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClientStore.java
raphkim: yep

PR #2387, opened 2021-03-19
chore(release): version bump

changxu0306:
Bump version from 2.22.2 to 2.22.3 for release.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -25,8 +25,8 @@ android.useAndroidX=true
 android.enableJetifier=true
 
 GROUP=com.amazonaws
-VERSION_NAME=2.22.2
-VERSION_CODE=022202
+VERSION_NAME=2.22.3
+VERSION_CODE=022203
 
 POM_URL=https://github.com/aws/aws-sdk-android
 POM_SCM_URL=https://github.com/aws/aws-sdk-android
--- a/gradle.properties
+++ b/gradle.properties
@@ -26,7 +26,6 @@ android.enableJetifier=true
 
 GROUP=com.amazonaws
 VERSION_NAME=2.22.3
-VERSION_CODE=022203
 
 POM_URL=https://github.com/aws/aws-sdk-android
 POM_SCM_URL=https://github.com/aws/aws-sdk-android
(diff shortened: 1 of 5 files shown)


gradle.properties
```diff
@@ -25,8 +25,8 @@ android.useAndroidX=true
 android.enableJetifier=true
 
 GROUP=com.amazonaws
-VERSION_NAME=2.22.2
-VERSION_CODE=022202
+VERSION_NAME=2.22.3
+VERSION_CODE=022203
```
richardmcclellan, line 29:
```suggestion
```

raphkim: @richardmcclellan, do we not use the version code anymore?

richardmcclellan: @raphkim we [removed it from amplify-android](https://github.com/aws-amplify/amplify-android/pull/1223), and the change was released yesterday.    My understanding is that it is not needed at all for an Android library - it's only needed for an Android app.

changxu0306: If we want to remove `VERSION_CODE`, I think it's better to put another PR.

richardmcclellan (commented):
Overall this looks :+1:, but unless these are urgent changes, could we wait until Monday morning to release?  Just to avoid any issues that may require attention over the weekend!

changxu0306:
> Overall this looks 👍, but unless these are urgent changes, could we wait until Monday morning to release? Just to avoid any issues that may require attention over the weekend!

Sure thing! Before that, I can also put a PR for removing `VERSION_CODE`.

gradle.properties
richardmcclellan: @changxu0306 okay! https://github.com/aws-amplify/aws-sdk-android/pull/2388

richardmcclellan: I just checked the [AppSync repo's `gradle.properties`](https://github.com/awslabs/aws-mobile-appsync-sdk-android/blob/main/gradle.properties), and it doesn't even have `VERSION_CODE`!    So, that further confirms removing it is fine.

raphkim (approved):

richardmcclellan (approved):

PR #2388, opened 2021-03-19
chore(release): remove VERSION_CODE

richardmcclellan:
Per the [documentation for `versionCode`](https://developer.android.com/studio/publish/versioning), it's only use is to prevent users from downgrading to an APK with a lower `versionCode` than the version currently installed.  Since `aws-sdk-android` is a library, and not an app, this doesn't apply.

We removed this from the [amplify-android library](https://github.com/aws-amplify/amplify-android/pull/1223) yesterday, and so far have had no issues reported from it.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,6 @@ subprojects { project ->
 private void configureAndroidLibrary(Project project) {
     project.android {
         defaultConfig {
-            versionCode rootProject.findProperty('VERSION_CODE').toInteger()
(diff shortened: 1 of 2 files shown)


rjuliano (approved):
🚢

changxu0306:
Should [this line](https://github.com/aws-amplify/aws-sdk-android/blob/main/build.gradle#L33) also be removed in `build.gradle(Project: aws-sdk-android)`?

PR #2386, opened 2021-03-18
register sign in providers for dropInUI

raphkim:
*Issue #, if available:*
#1003
#2368

*Description of changes:*
Call `registerConfigSignInProviders()` during initialization.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2393, opened 2021-03-24
prevent class def not found error

raphkim:
*Issue #, if available:*
Readdresses patch from #2386 

*Description of changes:*
Now that `registerConfigSignInProviders()` method is called (which has references to class paths in optional dependencies) in `_initialize()` step, `NoClassDefFoundError` would crash mobile client initialization even for those not using `showSignIn()`. 

This PR makes sure that initialization step will not crash for those using mobile client without optional dependencies.

### DIFF
--- a/aws-android-sdk-auth-core/src/main/java/com/amazonaws/mobile/auth/core/IdentityManager.java
+++ b/aws-android-sdk-auth-core/src/main/java/com/amazonaws/mobile/auth/core/IdentityManager.java
@@ -46,9 +46,8 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


aws-android-sdk-auth-core/src/main/java/com/amazonaws/mobile/auth/core/IdentityManager.java
```diff
@@ -122,8 +121,7 @@ private void setUnderlyingProvider(final CognitoCachingCredentialsProvider under
     private final CountDownLatch startupAuthTimeoutLatch = new CountDownLatch(1);
 
     /** Keep track of the registered sign-in providers. */
-    private final List<Class<? extends SignInProvider>> signInProviderClasses
-        = new LinkedList<Class<? extends SignInProvider>>();
+    private final Set<Class<? extends SignInProvider>> signInProviderClasses = new HashSet<>();
```
changxu0306, line 124: Why use `Set` to store `signInProviderClasses` instead of `List`?

raphkim: It removes the need to check for duplicates in `registerConfigSignInProviders()`. `getSignInProviderClasses()` returns a `Collection` anyways, so this is not a breaking change.

changxu0306 (approved):
Looks good!

rjuliano (approved):
ship it 🚢

richardmcclellan (commented):
I know this is already merged, but I just reviewed anyway, and it looks good!

PR #2415, opened 2021-04-03
feat: revoke tokens during auth sign out 

jamesonwilliams:
## What is this?

When sign out is requested with the `invalidateTokens(true)` option, `AWSMobileClient` will now revoke ID, access, and refresh tokens for the currently signed in Cognito User Pools user.

Tested with Drop-in UI and with native signIn/signOut flows. Not yet tested with Hosted UI.

Refer: https://github.com/aws-amplify/amplify-js/issues/3435

## A Note on Composition

The PR is broken up into two parts:
1. Updates to the Cognito Identity Provider client code, based on the services' update models.
2. Use of the new low-level client code from above, in the mid-level SDK clients and in the AWSMobileClient itself.

Testing:
- [x] HostedUI
- [x] User Pools
- [x] Federated Sign in
- [x] Unit tests


raphkim:
Just to note: codegen will override `service` package code if it sees any diff by default unless configured specifically to ignore certain files (in which case it will stop receiving updates from the service team).

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
```diff
@@ -2332,6 +2336,22 @@ private void deleteAttributesInternal(final List<String> attributeNamesToDelete,
         cognitoIdentityProviderClient.deleteUserAttributes(deleteUserAttributesRequest);
     }
 
+    public void revokeTokens() {
+        CognitoUserSession cognitoUserSession = getCachedSession();
+        String idToken = cognitoUserSession.getIdToken().getJWTToken();
+        if (!CognitoJWTParser.hasClaim(idToken, "jti")) {
```
lawmicha, line 2342: we are checking for "origin_jti" claim

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/services/cognitoidentityprovider/AmazonCognitoIdentityProviderClient.java
```diff
@@ -4534,6 +4534,57 @@ public InitiateAuthResult initiateAuth(InitiateAuthRequest initiateAuthRequest)
...
+     * @throws UnsupportedOperationException
+     * @throws AmazonClientException If any internal errors are encountered
+     *             inside the client while attempting to make the request or
+     *             handle the response. For example if a network connection is
+     *             not available.
+     * @throws AmazonServiceException If an error response is returned by Amazon
+     *             Cognito Your User Pool indicating either a problem with the
+     *             data in the request, or a server side issue.
+     */
+    public IntrospectTokenResult introspectToken(IntrospectTokenRequest introspectTokenRequest)
```
lawmicha, line 4559: this API is no longer available

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
```diff
@@ -2332,6 +2336,22 @@ private void deleteAttributesInternal(final List<String> attributeNamesToDelete,
         cognitoIdentityProviderClient.deleteUserAttributes(deleteUserAttributesRequest);
     }
 
+    public void revokeTokens() {
+        CognitoUserSession cognitoUserSession = getCachedSession();
+        String idToken = cognitoUserSession.getIdToken().getJWTToken();
```
lawmicha, line 2341: checking the accessToken instead of IdToken is enough

div5yesh:
PR for docs content change for token revocation: https://github.com/aws-amplify/docs/pull/3379

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
```diff
@@ -2332,6 +2334,24 @@ private void deleteAttributesInternal(final List<String> attributeNamesToDelete,
...
+        if (!CognitoJWTParser.hasClaim(accessToken, "origin_jti")) {
+            LOGGER.debug("Access Token does not contain `origin_jti` claim. Skip revoking tokens.");
+            return;
+        }
+        String refreshToken = cognitoUserSession.getRefreshToken().getToken();
+
+        RevokeTokenRequest request = new RevokeTokenRequest();
+        request.setToken(refreshToken);
+        request.setClientId(clientId);
+        if (!StringUtils.isBlank((clientSecret))) {
```
richardmcclellan, line 2349:
```suggestion
        if (!StringUtils.isBlank(clientSecret)) {
```

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
```diff
@@ -2332,6 +2334,24 @@ private void deleteAttributesInternal(final List<String> attributeNamesToDelete,
...
+            request.setClientSecret(clientSecret);
```
richardmcclellan, line 2350: Based on this logic, it looks revoking tokens will work whether you include `clientSecret` or not.  If that's the case, can we just not include `clientSecret` at all?

richardmcclellan: Ahh, nevermind.  Per [the docs](https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_RevokeToken.html#API_RevokeToken_RequestSyntax), it looks like it is required if the client id has a secret.

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -1611,6 +1611,9 @@ public Void run() throws Exception {
                     userpoolLL.globalSignOut(globalSignOutRequest);
                 }
                 if (signOutOptions.isInvalidateTokens()) {
+                    if (userpool != null) {
+                        userpool.getCurrentUser().revokeTokens();
+                    }
```
richardmcclellan, line 1616: If the revokeTokens call fails for any reason (poor internet connection, server error, etc.), will the custom tabs browser still be launched to attempt to clear the cookies?  I think we should.

Also, if revokeTokens fails, does the failure bubble up to the signout callback so the user knows that the tokens were not revoked?

div5yesh: Yes, if revokeTokens fails signout should continue. Revocation is best effort so a failure will be logged but the failure exception will be not be bubbled up.

richardmcclellan (approved):
LGTM!  🚀

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
```diff
@@ -2332,6 +2334,28 @@ private void deleteAttributesInternal(final List<String> attributeNamesToDelete,
         cognitoIdentityProviderClient.deleteUserAttributes(deleteUserAttributesRequest);
     }
 
+    public void revokeTokens() {
```
lawmicha, line 2337: on iOS PR, we decided to expose 2 APIs on the Cognito User. Please see this comment https://github.com/aws-amplify/aws-sdk-ios/pull/3638#discussion_r669191279  for more details as to the discussion that lead to this decision. I'm not sure if we need to align the APIs to this level of detail 
1. `isSessionRevocable` which returns bool for checking origin_jti claim 
2. `revokeToken` which is 1:1 with the LLC API and has a type that is from the codegenerated response types `AWSCognitoIdentityProviderRevokeTokenResponse`

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -1611,6 +1611,9 @@ public Void run() throws Exception {
...
                     if (hostedUI != null) {
```
lawmicha, line 1617: hostedUI sign out should also revoke the token, is this covered by this code path? ie. if hostedUI is not null, does that mean userpool is also not null

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -1611,6 +1611,9 @@ public Void run() throws Exception {
                     userpoolLL.globalSignOut(globalSignOutRequest);
                 }
                 if (signOutOptions.isInvalidateTokens()) {
+                    if (userpool != null) {
```
lawmicha, line 1614: isn't signOutOptions.invalidateTokens only used for hostedUI? what is the default value? if it is always false when options is not passed in, then the new behavior on line 1615 `userpool.getCurrentUser().revokeTokens();` will never be called. revoke tokens should always be called 
1. when user is signed in with user pools
2. when user is signed in with hostedUI and signOutOptions.invalidateTokens is true

revoke tokens should not be called for federated sign-ins

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
div5yesh: Yes, tokens are revoked for hostedUI sign out. `userpool` is alway non null whether hostedUI is null or not.

aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUser.java
div5yesh: > 1. `isSessionRevocable` which returns bool for checking origin_jti claim

The revokeToken returns either null or RevokeTokenResult. If revoke token fails due to origin_jti claim not found or server side error/exception, we just log the error and swallow the exception.

> 2. `revokeToken` which is 1:1 with the LLC API and has a type that is from the codegenerated response types `AWSCognitoIdentityProviderRevokeTokenResponse`

Added code changes.

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
div5yesh: Changed the default value of `invalidateTokens` to true. 

- [x] revoke tokens should always be called when user is signed in with user pools
- [x] revoke tokens should always be called when user is signed in with hostedUI and signOutOptions.invalidateTokens is true
- [ ] revoke tokens should not be called for federated sign-ins.

rjuliano: Maybe we need to move lines 1414-1416 outside of `if (signOutOptions.isInvalidateTokens()) {` and refactor the if statement to something like this:

```
if ((hostedUI != null && signOutOptions.isInvalidateTokens()) ||
      (hostedUI == null && userpool != null)) {
          userpool.getCurrentUser().revokeTokens();
      }
}
```
That way, you don't have to change the default value of `invalidateTokens` and we wouldn't have to worry a change in behavior. Let me know if this makes sense.

div5yesh: Turns out we don't need to change the default value of `invalidateTokens` since we have documented this in the SDK docs to do that manually. 

- User Pools
Amplify: Enforced by passing invalidateTokens(true) in AWSCognitoAuthPlugin.signOutLocally
SDK: Needs docs update similar to [SDK hostedUI](https://docs.amplify.aws/sdk/auth/hosted-ui/q/platform/android#sign-out-from-hostedui-1).

- HostedUI
Amplify: Enforced by passing invalidateTokens(true) in AWSCognitoAuthPlugin.signOutLocally
SDK: Manually enforced refer [docs](https://docs.amplify.aws/sdk/auth/hosted-ui/q/platform/android#sign-out-from-hostedui-1)

- Federation
Amplify: Only available through hostedUI.
SDK: Not enforced as expected. Users call `AWSMobileClient.getInstance().signOut()`.

rjuliano (commented):
LGTM 🚀

rjuliano (approved):

PR #2427, opened 2021-04-08
fix(mobile-client): missing optional dependency warning removed

raphkim:
*Issue #, if available:*

*Description of changes:*
removed warning from being logged when missing optional dependency. Those using `showSignIn()` are already expected to have these optional dependencies anyways and all it was doing was unnecessarily cluttering log messages for majority of users that do not need it.

### DIFF
--- a/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
+++ b/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
@@ -3544,8 +3544,7 @@ public final class AWSMobileClient implements AWSCredentialsProvider {
                 identityManager.addSignInProvider(GoogleSignInProvider.class);
             }
         } catch (NoClassDefFoundError exception) {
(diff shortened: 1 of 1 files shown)


aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -3544,8 +3544,7 @@ private void registerConfigSignInProviders(final AWSConfiguration awsConfigurati
                 identityManager.addSignInProvider(GoogleSignInProvider.class);
             }
         } catch (NoClassDefFoundError exception) {
-            Log.w(TAG, "Sign in provider was not registered due to missing optional dependency. " +
-                    "showSignIn() API may not work as expected.", exception);
+            // This is not 90% use-case. Do not log anything.
```
richardmcclellan, line 3547: Can you update this comment to explain the scenario that causes this exception to be thrown?  You could probably just put something similar to the log message that you are removing.

richardmcclellan (approved):

Issue #1947, opened 2020-06-26
Uploading large files to aws s3 almost always fails.


assassinshadow0:
**Describe the bug**
I am using the AWS S3 SDK to upload files from android. The files range from 128kb to 99 MB. Files that are less than ~20 MB gets uploaded without problems pretty easily. However, when attempting to upload larger files like 30-50MB or greater, I get time out error. 
Things I have tried -

**Editing ClientConfiguration** and setting a higher time out - Works on some phones but randomly gets timeout errors.
**Defining AmazonS3Client in application context** - Same results as above.
**Editing ClientConfiguration**  setting setSocketTimeout and setConnectionTimeout to 0 - This works ! however, the upload speed is not so good. Often times the upload just pauses for long moments (10 seconds to a whole minute give or take) and then resumes back up. 

**To Reproduce**
The code i am using -
` 
            CannedAccessControlList filePermission = CannedAccessControlList.PublicRead;

            TransferObserver observer = AppController.getInstance().getTransferUtility().upload(
                    "", //empty bucket name, included in endpoint
                    fileName,
                    new File(file), 

                    filePermission

            );

            observer.setTransferListener(new TransferListener() {
                @Override
                public void onStateChanged(int id, TransferState state)
                {

                    Log.e("TAG", state.name());

                    if (TransferState.COMPLETED.equals(state))
                    {

                        finishUp();

                    }
                    else if (TransferState.CANCELED.equals(state))
                    {
                        listener.onError(true);
                    }
                    else if (TransferState.FAILED.equals(state))
                    {

                        retryCount++;

                        if (retryCount >= 5)
                        {
                            listener.onError(true);
                        }
                        else
                        {
                            Handler handler = new Handler();
                            handler.postDelayed(new Runnable() {
                                @Override
                                public void run() {
                                    PerformUpload();
                                }
                            }, 5000);

                        }

                    }

                }

                @Override
                public void onProgressChanged(int id, long bytesCurrent, long bytesTotal) {
                    final double percent = (100. * (bytesCurrent / (double) bytesTotal));

                    mBuilder.setProgress(100, (int)percent, false);

                    if (retryCount > 0)
                    {

                        mBuilder.setContentTitle("Uploading..." + total + " Retry #" + retryCount);
                    }
                    else {
                        mBuilder.setContentTitle("Uploading..." + total);
                    }

                    mNotifyManager.notify(notif, mBuilder.build());
                }

                @Override
[...]

PR #2447, opened 2021-04-15
fix(s3): transfers are now single-threaded

raphkim:
*Issue #, if available:*
#1947 

*Description of changes:*
Default behavior for transfer utility was to use thread pool size = # available processors. This allowed multipart upload to be executed in parallel for a minor performance boost. However, we found that this was causing certain upload part requests to be prioritized over others, which in turn timed out the requests that were not being prioritized.

To increase the stability of large file uploads, this PR aims to force single-threaded multipart uploads by default (parts will be uploaded serially).

Main task executor to be used for downloads and single-part uploads will still be multi-threaded so that small transfers are not blocked by large file uploads.

### DIFF
--- a/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferThreadPool.java
+++ b/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferThreadPool.java
@@ -41,7 +41,8 @@ class TransferThreadPool {
             executorMainTask = buildExecutor(poolSize);
         }
         if (executorPartTask == null) {
-            executorPartTask = buildExecutor(poolSize);
+            // Upload individual parts serially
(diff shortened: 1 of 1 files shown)


richardmcclellan:
Could we just change the default pool size to 1, defined [here](https://github.com/aws-amplify/aws-sdk-android/blob/8de48797ae2e5c157169a94746c768da7de61056/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferUtilityOptions.java#L162) instead?  That way, we won't break the override.

raphkim:
Main task executor was left untouched intentionally so that one large file upload does not block other transfers completely. If a user wants to upload a 1GB file while downloading 3 small files (5KB each) from S3, then the downloads will not start until upload is completed.

richardmcclellan (approved):

PR #2448, opened 2021-04-16
change the time offset precision from int to long

drochetti:
**Notes:**

The clock skew auto-correct logic in the SDK relies on the `int` primitive type when calculating the offset. When the offset is converted from milliseconds to days, the ms represented as an `int` have the boundaries as -24 and +24 days. Changing it to long (64-bit precision) fixes the limit.

---

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/AmazonWebServiceClient.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/AmazonWebServiceClient.java
@@ -95,7 +95,7 @@ public abstract class AmazonWebServiceClient {
 
     /** Optional offset (in seconds) to use when signing requests. */
     @SuppressWarnings("checkstyle:visibilitymodifier")
-    protected int timeOffset;
+    protected long timeOffset;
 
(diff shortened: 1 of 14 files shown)


rjuliano (commented):
The change itself looks good to me. I do have one concern though.

I was looking at the Java SDK repo and noticed that they also use int for the offset. There is actually [code in place](https://github.com/aws/aws-sdk-java-v2/blob/master/core/sdk-core/src/main/java/software/amazon/awssdk/core/internal/retry/ClockSkewAdjuster.java#L54) to enforce this.

They're passing the skew (in seconds) into the `Math.toIntExact` function, which appears to [throw an `ArithmeticException`](https://docs.oracle.com/javase/8/docs/api/java/lang/Math.html#toIntExact-long-) if the input value overflows the max value of an int. If that happens, there's a warning logged and the the skew calculation defaults to 0.

You have far more background on this than I do, but I just wanted to bring that up for discussion.

raphkim (approved):
The code changes look fine as long as we confirm that not enforcing `int` doesn't break anything as RafJ pointed out.

PR #2395, opened 2021-03-25
feat(aws-android-sdk-cognitoidentityprovider): support custom endpoint

thisisabhash:
*Issue #, if available:* NA

*Description of changes:* Support for Custom endpoints in Cognito User Identity Pool

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2455, opened 2021-04-21
feat(aws-android-sdk-cognitoidentityprovider): support custom endpoint

thisisabhash:
This reverts commit b783ffacddb7328d217ac3d0689edfe4221b2ebe.

*Issue #, if available:* None

*Description of changes:* Code changes to support custom end points for Cognito User Pools.

### DIFF
--- a/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUserPool.java
+++ b/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/CognitoUserPool.java
@@ -274,7 +274,6 @@ public class CognitoUserPool {
         this.pinpointEndpointId = CognitoPinpointSharedContext.getPinpointEndpoint(context, pinpointAppId);
     }
 
-
     /**
      * Constructs a user-pool with default {@link ClientConfiguration}.
      *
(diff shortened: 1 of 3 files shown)


richardmcclellan:
It looks like this is an exact resubmission of https://github.com/aws-amplify/aws-sdk-android/pull/2395, which was reverted in https://github.com/aws-amplify/aws-sdk-android/pull/2425.   So, it sounds like the only difference is that you've completed manual testing and confirmed that the original PR is actually safe, is that right?    Could you briefly describe the manual testing steps that you took to verify this change?

thisisabhash:
> It looks like this is an exact resubmission of #2395, which was reverted in #2425. So, I'm guessing the only difference is that you've completed manual testing and confirmed that the original PR is actually safe, is that right? Could you briefly describe the manual testing steps that you took to verify this change?

Yes, that's correct. I did manual testing by creating a sample app through Amplify Android and referencing the local SDK with my changes in it.

- I have tested the following APIs of AuthCategory with authentication flow type as USER_SRP_AUTH and CUSTOM_AUTH.
signUp, confirmSignUp, confirmSignIn, signOut, fetchAuthSession, fetchUserAttributes, updatePassword, resetPassword, confirmResetPassword, rememberDevice, forgetDevice, fetchDevices, updateUserAttribute, getCurrentUser

- To check for failure, I gave a dummy custom end point which lead to an exception [AmazonClientException: Unable to execute HTTP request: Unable to resolve host "xxxx.xxx.xxx": No address associated with hostname]

- To check for REFRESH_TOKEN_AUTH flow, I signed in and waited for tokens to expire. Then I invoked fetchAuthSession to trigger the REFRESH_TOKEN_AUTH flow and verified that the endpoint for this request was the custom endpoint.

- To check for MFA authentication, I tested the signUp, confirmSignUp, signIn and confirmSignIn by receiving a sign in code as an SMS and using it to sign in.

richardmcclellan (approved):
Thanks for the explanation @thisisabhash !  LGTM!

PR #2458, opened 2021-04-22
fix(pinpoint): add campaign attributes to push events

raphkim:
*Description of changes:*
Journey implementation assumed that journey and campaign events were mutually exclusive and incorrectly removed campaign attributes even when they were expected in the payload. This PR reverses that change.


changxu0306 (approved):

richardmcclellan:
Is this fixing a regression introduced by https://github.com/aws-amplify/aws-sdk-android/pull/1542/?    I was looking at the changes in that PR, and it doesn't seem like this is reverting anything?  Kinda hard to follow the previous PR though.  Could you just explain how this fix works?

raphkim:
#1542 renamed `setCampaignAttributes()` method to `updateEventSourceGlobally()` and put that method inside the `if-else` statement that checks whether the app is in foreground or not (thereby adding the attributes only if the app is in the background). This PR calls the new `updateEventSourceGlobally()` unconditionally again.

richardmcclellan (approved):

PR #2472, opened 2021-05-10
fix: launch hosted-ui sign-out using custom tabs manager

raphkim:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/1090
https://github.com/aws-amplify/amplify-android/issues/1101

*Description of changes:*
Sign out flow was being launched without using `CustomTabsManagerActivity` which would cause `CustomTabsRedirectActivity` to be involved at an incorrect state. This would launch `CustomTabsManagerActivity` from the middle of the flow instead from the beginning.

This PR starts the flow correctly so that the entire auth flow can be carried out as intended.

Note: this PR does not address the fact that sign-out callback is not working as intended, which is a different issue.


changxu0306 (approved):
Ship it 🚢

PR #2499, opened 2021-06-02
fix(mobileclient): Honor auth flow setting from config

rjuliano:
*Issue #, if available:* #2349

*Description of changes:*
Another attempt at fixing the above issue. The problem with the last fix is that we were mixing auth flow type names with the challenge name that should be set in AuthenticationDetails.

### DIFF
--- a/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
+++ b/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
@@ -29,6 +29,7 @@ import android.net.Uri;
 import android.os.Build;
 import android.util.Log;
 import androidx.annotation.AnyThread;
+import androidx.annotation.VisibleForTesting;
 import androidx.annotation.WorkerThread;
 import androidx.browser.customtabs.CustomTabsCallback;
 import androidx.browser.customtabs.CustomTabsClient;
@@ -129,6 +130,9 @@ import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 
 import static com.amazonaws.mobile.client.results.SignInState.CUSTOM_CHALLENGE;
+import static com.amazonaws.mobileconnectors.cognitoidentityprovider.util.CognitoServiceConstants.AUTH_TYPE_INIT_CUSTOM_AUTH;
+import static com.amazonaws.mobileconnectors.cognitoidentityprovider.util.CognitoServiceConstants.AUTH_TYPE_INIT_USER_PASSWORD;
+import static com.amazonaws.mobileconnectors.cognitoidentityprovider.util.CognitoServiceConstants.CHLG_TYPE_USER_PASSWORD;
 
 /**
  * The AWSMobileClient provides client APIs and building blocks for developers who want to create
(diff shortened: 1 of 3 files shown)


aws-android-sdk-mobile-client/src/test/java/com/amazonaws/mobile/client/AWSMobileClientAuthFlowSettingTest.java
```diff
@@ -0,0 +1,197 @@
+package com.amazonaws.mobile.client;
```
raphkim, line 1: copyright :)

this is great, though

aws-android-sdk-mobile-client/src/test/java/com/amazonaws/mobile/client/AWSMobileClientAuthFlowSettingTest.java
```diff
@@ -0,0 +1,197 @@
...
+@RunWith(RobolectricTestRunner.class)
+public class AWSMobileClientAuthFlowSettingTest {
+    private static final String TAG = AWSMobileClientAuthFlowSettingTest.class.getSimpleName();
+    CountDownLatch initLatch;
+    CountDownLatch signinLatch;
+    CountDownLatch continuationLatch;
+    AtomicBoolean initComplete;
+    AtomicBoolean signInComplete;
+    AtomicBoolean continuationComplete;
+    AWSMobileClient mobileClient;
```
raphkim, line 46: can't these be private?

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -1213,23 +1236,40 @@ public void onSuccess(CognitoUserSession userSession, CognitoDevice newDevice) {
...
                                         } else {
                                             authenticationContinuation.setAuthenticationDetails(new AuthenticationDetails(username, authParameters, validationData));
                                         }
+                                    } else if (authFlowTypeInConfig && AUTH_TYPE_INIT_USER_PASSWORD.equals(authFlowType)) {
+                                        // If there's a value in the config and it's USER_PASSWORD_AUTH, set the auth type (challenge name)
+                                        // to be USER_PASSWORD.
+                                        AuthenticationDetails authenticationDetails = new AuthenticationDetails(username, password, validationData);
+                                        authenticationDetails.setAuthenticationType(CHLG_TYPE_USER_PASSWORD);
+                                        authenticationContinuation.setAuthenticationDetails(authenticationDetails);
+
```
raphkim, line 1263: is this the only branch that's being added from the original logic?

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -346,6 +350,25 @@ public static synchronized AWSMobileClient getInstance() {
...
+        if (forceGetNew) {
+            singleton = null;
+        }
+        return new AWSMobileClient();
+    }
+
+    @VisibleForTesting
+    void setUserPool(CognitoUserPool userpool) {
+        this.userpool = userpool;
+    }
```
raphkim, line 370: would it be possible to inject a mock userpool during initialization by adding something like
```
"CognitoUserPool": {
    "Default": {
        "PoolId": "<pool-id>",
        "AppClientId": "<client-id>",
        "Region": "<region>"
    }
},
```
during config generation? That would be preferable to adding more `@VisibleForTesting` methods.

raphkim (approved):
Just wondering if we can remove unnecessary `@VisibleForTesting` methods, but looks great otherwise. Thanks for your work!

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
rjuliano: It would be tricky because you actually have to inject the mock after the initialize method runs. It would probably require a lot more refactoring, so I'm not sure it would be worth it.

PR #2503, opened 2021-06-07
Read the `authenticationFlowType` value of the configuration

### LINKED ISSUE #1488: Enable custom auth flow on demand
**State your question**
Our default configuration in the config file is set to use `OAuth`, we use this to handle most of the authentication flows on our app, however we provide a password less flow that needs to be handled using custom auth. The problem is that if I initialise  the client **with** custom auth  flows enabled, regular `OAuth` flows won't work and if I initialise the client **with out** custom auth our password less flow always triggers and exception:

`NoAuthorizedException: Incorrect username or password`

My question then is: is possible to enable/disable custom auth flow on demand and if so, how? Or re-initialise the client so it will load the desired config?

**Which AWS Services are you utilizing?**
Cognito User Pools through the AWSMobileClient

**Device Information (please complete the following information):**
 - Device: Pixel 3
 - Android Version: Android 10
 - Specific to simulators: No

**Environment (please complete the following information):**
SDK Version: 2.16.7

bogren:
Attempt to fix #1488. Related to but doesn't fix https://github.com/aws-amplify/aws-sdk-android/issues/684. A small step to show the need for customizing programmatically after the SDK has been initialized. 

The change is inspired from the workaround described in https://github.com/aws-amplify/aws-sdk-ios/issues/1204 and the original PR that looks slightly different when it's initially introduced on iOS https://github.com/aws-amplify/aws-sdk-ios/pull/1860.

The idea with the change is to allow us to programmatically change to custom auth before attempting login, calling the custom challenge, then revert the change after the sign in, if we fail the user could change/try a different method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## [Release 2.25.0](https://github.com/aws-amplify/aws-sdk-android/releases/tag/release_v2.25.0)
+
+### Features
+- **aws-android-sdk-cognitoidentityprovider:** update models to latest (#2510)
+
+### Bug Fixes
+- **aws-android-sdk-lex:** prioritize custom lex signer for all regions (#2506)
+- **mobileclient:** Honor auth flow setting from config (#2499)
+- **aws-android-sdk-polly:** use correct SignerConfig in all regions (#2505)
+
+[See all changes between 2.24.0 and 2.25.0](https://github.com/aws-amplify/aws-sdk-android/compare/release_v2.24.0...release_v2.25.0)
+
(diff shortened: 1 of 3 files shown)


raphkim (commented):
Hi @bogren! Thank you for your contribution to the project.
We do not officially support the use-case of programmatically switching between different auth flow types at the moment, but I can see how this PR can provide a potential workaround. We will investigate to see if there are any unintended side effects of this and provide an update.

bogren:
Hi, yes I understand. Thanks, appreciate you looking into the possibilities to get a workaround like this merged.

PR #2504, opened 2021-06-08
fix(aws-android-sdk-s3): implement retry mechanism for upload part

raphkim:
*Issue #, if available:*

*Description of changes:*


rjuliano (approved):
LGTM 🚀

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/UploadPartTask.java
```diff
@@ -53,53 +55,75 @@ public UploadPartTask(UploadTask.UploadPartTaskMetadata uploadPartTaskMetadata,
...
+                final UploadPartResult putPartResult = s3.uploadPart(uploadPartRequest);
+                setTaskState(TransferState.PART_COMPLETED);
+                dbUtil.updateETag(uploadPartRequest.getId(), putPartResult.getETag());
+                return true;
+            } catch (AbortedException e) {
+                // If request got aborted, operation was paused or canceled. do not retry.
+                LOGGER.debug("Upload part aborted.");
+                resetProgress();
+                return false;
+            } catch (final Exception e) {
```
richardmcclellan, line 72: It looks like this is going to retry for any error.   Ideally, we only want to retry for errors that are retryable (e.g. TimeoutException, any 5xx error), and not for errors that we know will just fail (Unauthorized errors, any 4xx error, etc.)

Can this be updated so that we don't retry for 4xx errors?

richardmcclellan: As discussed offline, any 4xx errors will be caught on the request that initiates the multipart upload.  These are not expected from the requests which upload the parts.

PR #2505, opened 2021-06-08
fix(polly): use correct SignerConfig in all regions

tllauda:
*Issue #, if available:*

*Description of changes:*
Polly should use `AmazonPollyCustomPresigner` in all regions. Currently we're receiving `403` in `eu-central-1`.


raphkim (approved):
Thank you for this fix! This also addressed a similar issue in Lex.

tllauda:
The fix is included in 2.25.0 release: https://github.com/aws-amplify/aws-sdk-android/releases/tag/release_v2.25.0.

Issue #2485, opened 2021-05-19
Error: The request signature we calculated does not match the signature you provided!! | LEXSAMPLE


HaemanthSP:
*LexSample*: I am facing the signature mis-match error after upgrading to aws version: 2.23.0 or (2.22.6 +) as indicated in #2385. 

**To Reproduce**
Upgrade to AWS version: 2.22.6 or later

Which requires androidx migration: After migration dependencies looks like this
```
dependencies {
    def android_libs_version = "28.0.0"
    implementation 'androidx.multidex:multidex:2.0.0'
    implementation 'androidx.appcompat:appcompat:1.0.0'
    implementation 'com.google.android.material:material:1.0.0'

    def aws_version = '2.23.0'
    implementation ("com.amazonaws:aws-android-sdk-lex:$aws_version@aar") { transitive=true }
    implementation ("com.amazonaws:aws-android-sdk-mobile-client:$aws_version@aar") { transitive=true }
}
```
**Which AWS service(s) are affected?**

AmazonLexRuntimeClient

**Expected behavior**

During the start of the conversation, encountered following error

```
E/TextActivity: Interaction error
    com.amazonaws.AmazonServiceException: The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details. (Service: AmazonLexRuntime; Status Code: 403; Error Code: InvalidSignatureException; Request ID:...
        at com.amazonaws.http.AmazonHttpClient.handleErrorResponse(AmazonHttpClient.java:742)
        at com.amazonaws.http.AmazonHttpClient.executeHelper(AmazonHttpClient.java:420)
        at com.amazonaws.http.AmazonHttpClient.execute(AmazonHttpClient.java:229)
        at com.amazonaws.services.lexrts.AmazonLexRuntimeClient.invoke(AmazonLexRuntimeClient.java:885)
        at com.amazonaws.services.lexrts.AmazonLexRuntimeClient.postContent(AmazonLexRuntimeClient.java:623)
        at com.amazonaws.mobileconnectors.lex.interactionkit.InteractionClient$3.run(InteractionClient.java:452)
        at java.lang.Thread.run(Thread.java:764)
```

**Environment Information (please complete the following information):**
 - AWS Android SDK Version: 2.22.6 and above
 - Device: Both in simulator and actual device
 - Android Version: Android 8.0
 - Specific to simulators: No

**Additional context**
[...]

Issue #2385, opened 2021-03-18
Error: credential should be scoped to the correct service lex


cicchello82:
Hi,
I'm trying the LexSample. I followed all configurations steps but if I launch the App and try to send text command to bot, I have the error **credential should be scoped to the correct service lex**
Anyne can help me?
Thanks in advance,
Giacomo

raphkim:
Hi @cicchello82 ,
As per our last discussion in Discord, I was unable to reproduce this issue in the LexSample app. Would you be able to update this ticket with more information to help us find out what may be causing this issue for you? The rest of the stacktrace and detailed configuration steps that you took (without any sensitive information) would be helpful.

cicchello82:
Hi @raphkim,
I tried two steps: first I followed the instruction in readme file for installing LexSample and it gave me the error.
The I import the sample in a working Android app with amplify Auth and Analyticsm but I had same error when I try to call the Bot.
I don't now if I miss some configuration settings. Could you please me wich part of code show you to better understand?
Thanks very much,
Giacomo

richardmcclellan:
We believe this issue was fixed in the [2.22.6 release](https://github.com/aws-amplify/aws-sdk-android/releases/tag/release_v2.22.6).  The latest version is [2.22.7](https://github.com/aws-amplify/aws-sdk-android/releases/tag/release_v2.22.7) - please give that a try!  If you are still having issues, feel free to comment or open a new issue.

cicchello82:
Hi,
after upgrading libraries to 2.22.7 something changed but I have a new error "the request signature we calculated does not match the signature you provided. Check your aws secret.....".

HaemanthSP:
Hi @cicchello82 ,

Have you found the solution?
I am facing the same issue.

>  com.amazonaws.AmazonServiceException: The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.

Best regards,
Haemanth

PR #2506, opened 2021-06-08
fix(aws-android-sdk-lex): prioritize custom lex signer for all regions

raphkim:
*Issue #, if available:*
#2385 
#2485

*Description of changes:*
Default sigv4 signer was being used for lex service in `eu-central-1` and `cn-north-1` regions. Now Lex custom signer is used for both regions too.


rjuliano (approved):
LGTM 🚀

PR #2529, opened 2021-06-28
fix(aws-android-sdk-cognitoauth): custom tabs app crash when chrome not available

div5yesh:
- add support for launching intent chooser for custom tabs
 - fix app crash if chrome not available by choosing availalbe browser

*Issue #, if available:*

*Description of changes:*


aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
```diff
@@ -709,21 +713,56 @@ private void launchSignOut(final String redirectUri, final String browserPackage
...
      */
     private void launchCustomTabs(final Uri uri, final Activity activity, final String browserPackage) {
     	try {
 	        CustomTabsIntent.Builder builder = new CustomTabsIntent.Builder(mCustomTabsSession);
 	        mCustomTabsIntent = builder.build();
 	        if(pool.getCustomTabExtras() != null)
 	            mCustomTabsIntent.intent.putExtras(pool.getCustomTabExtras());
-	        mCustomTabsIntent.intent.setPackage(
-	                browserPackage != null ? browserPackage : DEFAULT_BROWSER_PACKAGE);
+	        if(browserPackage != "customtabs.intent.chooser") {
```
rjuliano, line 762: If browserPackage is equals to `customtabs.intent.chooser`, where is it actually getting launched?

div5yesh: If browserPackage is equal to `customtabs.intent.chooser`, the url will open in default browser. But if the default browser is not available, it will launch intent chooser. For intent chooser to launch, `CustomTabsIntent` package should be null.

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
```diff
@@ -709,21 +713,56 @@ private void launchSignOut(final String redirectUri, final String browserPackage
...
+	            mCustomTabsIntent.intent.setPackage(
```
richardmcclellan, line 763: Do we even need to call `setPackage` at all?  What happens if we don't set it?  Per this custom tabs integration guide, it doesn't actually say you need to set it.   Perhaps if we don't set it, then it will just proceed if one is present, or prompt the user if there are multiple?    Prompting the user if multiple are installed is really the ideal behavior anyway.

div5yesh: Yes that's correct. If we don't set the package, it will just use the one set as default. If default is not set then use the browser that is available or prompt the user in case of multiple.

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
div5yesh: Removed `customtabs.intent.chooser`. We will not set any default packages.

richardmcclellan (commented):
These changes are so simple!   :)   It seems promising.  A couple questions:

 - What happens now if there is no browser installed that supports custom tabs?
 - What happens if there is no browser installed at all?
 - What happens if there are two or more browsers installed that support custom tabs?
 - Will this ever result in a launching a browser that does not support custom tabs?

div5yesh:
> * What happens now if there is no browser installed that supports custom tabs?

It will launch in a standalone browser. https://developer.chrome.com/docs/android/custom-tabs/best-practices/#provide-a-fallback-for-when-custom-tabs-is-not-installed

> * What happens if there is no browser installed at all?

It will result in an exception:

```java
java.lang.RuntimeException: Unable to resume activity {com.example.customtabs/com.amazonaws.mobileconnectors.cognitoauth.activities.CustomTabsManagerActivity}: android.content.ActivityNotFoundException: No Activity found to handle Intent
```

> * What happens if there are two or more browsers installed that support custom tabs?

It will launch in the system browser or the user’s default browser. If no default is set, it will prompt the intent chooser.

> * Will this ever result in a launching a browser that does not support custom tabs?

Yes, it will always launch in a standalone browser(if available) when custom tabs is not supported.

Refer: https://developer.chrome.com/docs/android/custom-tabs/integration-guide/#what-happens-if-the-user-doesnt-have-a-browser-that-supports-custom-tabs-installed

div5yesh:
Docs PR to include queries section into manifest file: https://github.com/aws-amplify/docs/pull/3401

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
```diff
@@ -709,21 +713,68 @@ private void launchSignOut(final String redirectUri, final String browserPackage
...
+        return packageNamesSupportingCustomTabs;
+    }
+
     /**
      * Launches the HostedUI webpage on a Custom Tab.
      * @param uri Required: {@link Uri}.
      * @param activity Activity to launch custom tabs from and which will listen for the intent completion.
      * @param browserPackage Optional string specifying the browser package to launch the specified url.
-     *                       Defaults to Chrome if null.
+     *                       Launches intent chooser if set to null.
```
raphkim, line 759: what happens if it is null and customer only has one browser (e.g. chrome)? Would it default to that one without prompting?

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
```diff
@@ -709,21 +713,68 @@ private void launchSignOut(final String redirectUri, final String browserPackage
...
 	        CustomTabsIntent.Builder builder = new CustomTabsIntent.Builder(mCustomTabsSession);
 	        mCustomTabsIntent = builder.build();
 	        if(pool.getCustomTabExtras() != null)
 	            mCustomTabsIntent.intent.putExtras(pool.getCustomTabExtras());
-	        mCustomTabsIntent.intent.setPackage(
-	                browserPackage != null ? browserPackage : DEFAULT_BROWSER_PACKAGE);
+	        if(browserPackage != null) {
+	            mCustomTabsIntent.intent.setPackage(browserPackage);
+            }
             mCustomTabsIntent.intent.setData(uri);
```
raphkim, line 778: whitespaces looks a bit wonky here. can we get it addressed?

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
```diff
@@ -709,21 +713,68 @@ private void launchSignOut(final String redirectUri, final String browserPackage
         launchCustomTabs(fqdn, null, browserPackage);
     }
 
+    /***
+     * Check if a browser is installed on the device to launch HostedUI.
+     * @return true if a browser exists else false.
+     */
+    public Boolean isBrowserInstalled() {
```
raphkim, line 720: can we make this `private` unless it is being accessed elsewhere?

raphkim, line 720: This method should cache the result so that we aren't resolving this activity more than once.
Refer to similar logic being handled [here](https://github.com/aws-amplify/aws-sdk-android/blob/74209c4914f11a0476025e0d656e9182e56d1506/aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java#L775-L778).

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
```diff
@@ -709,21 +713,68 @@ private void launchSignOut(final String redirectUri, final String browserPackage
...
+        Uri webAddress = Uri.parse(url);
+        Intent intentWeb = new Intent(Intent.ACTION_VIEW, webAddress);
+        return (intentWeb.resolveActivity(context.getPackageManager()) != null);
+    }
+
+    /**
+     * Get list of browser packages that support Custom Tabs Service.
+     * @return list of package names that support Custom Tabs.
+     */
+    private ArrayList<String> getSupportedBrowserPackage(){
```
raphkim, line 731: we should cache the result for this as well

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
raphkim: We should return primitive type `boolean` here rather than `Boolean` object, because the former is cheaper and guaranteed to be non-null.

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
```diff
@@ -709,21 +713,68 @@ private void launchSignOut(final String redirectUri, final String browserPackage
...
      * @param uri Required: {@link Uri}.
      * @param activity Activity to launch custom tabs from and which will listen for the intent completion.
      * @param browserPackage Optional string specifying the browser package to launch the specified url.
-     *                       Defaults to Chrome if null.
+     *                       Launches intent chooser if set to null.
      */
     private void launchCustomTabs(final Uri uri, final Activity activity, final String browserPackage) {
     	try {
+    	    if(!isBrowserInstalled()) {
+                userHandler.onFailure(new Exception("No browsers installed."));
```
raphkim, line 764: We should use `RuntimeException` to be (slightly) more specific. Same for when custom tabs browser isn't found. It is generally best practice to use as specific of an exception type as possible (even if it isn't followed throughout this project unfortunately).

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
raphkim, line 731: It is usually best to return a more generic type such as `List<String` or `Collection<String>` depending on how the method is being used. Since we are only using the result of this method to check for the size, `Collection<String>` would make most sense here.

raphkim: Actually, why cant this method simply be `isCustomTabsSupported()` and return a `boolean`? Are we using the list of supported packages elsewhere?

raphkim (changes requested):

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
div5yesh: Yes, it will default to that browser instead of the prompt.

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
div5yesh: I have wrapped `getSupportedBrowserPackage` into `isCustomTabsSupported()`. Although we are not using the list, kept it just to be more readable.

raphkim (approved):
🚢

aws-android-sdk-cognitoauth/src/main/java/com/amazonaws/mobileconnectors/cognitoauth/AuthClient.java
```diff
@@ -709,21 +728,90 @@ private void launchSignOut(final String redirectUri, final String browserPackage
         launchCustomTabs(fqdn, null, browserPackage);
     }
 
+    /***
+     * Check if a browser is installed on the device to launch HostedUI.
+     * @return true if a browser exists else false.
+     */
+    private boolean isBrowserInstalled() {
+        if (isBrowserInstalled) {
+            return true;
```
richardmcclellan, line 737: Looks like this caches the value if it is true, but not if it is false.  Can we cache it for false as well?  You could make it a `Boolean` instead of `boolean`, and check for null instead of checking if it's true.

div5yesh: The value is set false by default. We should query for browsers every time in case the user installs a browsers after a failure. If we cache false value then the custom tabs won't launch even if the user just installed a browser. It will work though if user restarts the app.

PR #2546, opened 2021-07-19
Changes to support encoded resource path coming in request.

poojamat:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/Request.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/Request.java
@@ -268,12 +268,12 @@ public interface Request<T> {
     /**
      * Get Encoded Uri Resource Path
      */
-    String getEncodedUriResourcePath();
+    public String getEncodedUriResourcePath();
 
     /**
      *
      * Set Encoded Uri Resource Path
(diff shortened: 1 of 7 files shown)


aws-android-sdk-core/src/main/java/com/amazonaws/Request.java
```diff
@@ -61,13 +61,15 @@
      *
      * @param path The path to the resource being requested.
      */
+    @Deprecated
```
richardmcclellan, line 64: I don't think we should mark this as deprecated, since we still plan to rely on this method for the majority of the existing SDKs.

poojamat: This method shouldn't be used for newer implementation though. Otherwise we will be struggling with encoding issues.

richardmcclellan: As discussed, it's really just a question of whether we want a warning to show up.  Thinking about it some more, I do think it's a good idea to mark the old ones as deprecated, so that when codegen adds support for new SDKs, they opt for the new approach.

richardmcclellan (approved):
🚀

PR #2547, opened 2021-07-20
fix: add overload of confirmResetPassword with user

rjuliano:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-android/issues/1363

*Description of changes:*
This is a feature parity issue that was brought up by the Flutter team as well as in the issue referenced above.

If a user begins the reset password flow, but exits the app before entering the confirmation code, the user is unable to complete the reset password process. iOS handles this by allowing the caller to pass in the user name as one of the parameters.

We don't have a corresponding API in AWS mobile client that would allow us to add such an overload to `amplify-android`. The intent of this PR is to fill that gap by re-creating the `ForgotPasswordContinuation` object. One concern I have is that once the session is done, we no longer have the data to populate the `CognitoUserCodeDeliveryDetails` which is passed in as a parameter to the `ForgotPasswordContinuation` constructor. In this PR I'm passing null, since as best as I can tell, that data doesn't seem to be used downstream.


haverchuck:
I haven't tested, but this looks like what we need.

PR #2555, opened 2021-07-28
feat(aws-android-sdk-mobile-client): adds getUserSub method

haverchuck:
*Issue #, if available:*
Assist amplify-flutter in resolving a discrepancy between android and ios behavior.

*Description of changes:*
Adds a getUserSub method to the mobileclient; similar to getUsername.


aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
```diff
@@ -844,6 +845,25 @@ public String getUsername() {
...
+     * @return The sub attribute of the current access token.
+     * @see <a href="https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-with-identity-providers.html#amazon-cognito-user-pools-using-the-access-token">Using the Access Token</a>
+     * from Cognito documentation.
+     */
+    @AnyThread
+    public String getUserSub() {
+        try {
+            if (userpoolsLoginKey.equals(mStore.get(PROVIDER_KEY))) {
+                String token = mStore.get("token");
+                return (String) CognitoJWTParser.getPayload(token).get("sub");
```
raphkim, line 859: is this equivalent to the following?
```
return CognitoJWTParser.getClaim(token, "sub");
```

raphkim (commented):
Can't the customers simply extract the appropriate claim out of `AWSMobileClient#getTokens()` so that they can guarantee that the session is refreshed and vended for the current user?

Or at least, the implementation of `getUserSub()` should make use of that logic internally.

aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java
haverchuck: I think so? I'm not 100% sure.

raphkim: they do the [exactly same thing](https://github.com/aws-amplify/aws-sdk-android/blob/1940e0500f50e34b5096e4635a4b60beb0fa8284/aws-android-sdk-cognitoidentityprovider/src/main/java/com/amazonaws/mobileconnectors/cognitoidentityprovider/util/CognitoJWTParser.java#L105-L106) :) I was only suggesting it because it's much cleaner and this is exactly what it's supposed to be used for

rjuliano (approved):
LGTM 🚀

haverchuck:
@raphkim I incorporated your suggestion.

raphkim:
I am concerned about how this PR is seeking to vend out authenticated user's information from the cache to an unauthenticated user. This behavior should be reviewed across platforms before making it to main branch.

raphkim (changes requested):
I am concerned about how this PR is seeking to vend out authenticated user's information from the cache to an unauthenticated user. This behavior should be reviewed across platforms before making it to main branch.

raphkim (approved):
The potential security concern has been dismissed in an offline discussion 🚢

Issue #2558, opened 2021-08-03
AWSlotMqttManager offlinePublishQueueBound not working


CapitainOats:
**Describe the bug**
The `offlinePublishQueueBound` in the `AWSlotMqttManager` fails to apply an actual message limit. In the method `putMessageInQueueAndNotify` is called `mqttMessageQueue.remove(0);` which intends to remove first item of a queue, however `ConcurentLinkedQueue.remove` accepts only an Object of given queue as a parameter. This results into the remove call never removing items.

**Which AWS service(s) are affected?**
aws-android-sdk-iot

raphkim:
Thank you for catching this issue @CapitainOats! I have a PR to address the bug and we will close the issue once it is released.

CapitainOats:
@raphkim Thanks for the fast response. I closed the duplicated PR #2559 created by me.

PR #2559, opened 2021-08-03
fix(aws-android-sdk-io): AWSlotMqttManager offlinePublishQueueBound n…

CapitainOats:
*Issue #, if available:*
#2558 

*Description of changes:*
Calling the ConcurrentLinkedQueue.remove(0) would result into removing
Integer initialized to 0 from an Integer queue, however in this case it
always fails to remove first AWSIotMqttQueueMessage Object.

Switch the unsupported ConcurrentLinkedQueue.remove(0) call to
the supported poll.


CapitainOats:
Closing because of duplicity with the PR #2565.

Issue #2560, opened 2021-08-04
Wrong null check for signInUIOptions of AWSMobileClient#showSignIn


monae:
**Describe the bug**
AWSMobileClient#showSignIn will be failed with internal exception when certain signInUIOptions is given.
In the code below, null check should be applied to `hostedUIOptions.getTokenQueryParameters()`, not `hostedUIOptions.getSignInQueryParameters()`.

https://github.com/aws-amplify/aws-sdk-android/blob/release_v2.27.0/aws-android-sdk-mobile-client/src/main/java/com/amazonaws/mobile/client/AWSMobileClient.java#L3203-L3207
```
if (hostedUIOptions.getSignInQueryParameters() != null) {
    for (Map.Entry<String, String> e : hostedUIOptions.getTokenQueryParameters().entrySet()) {
        tokensUriBuilder.appendQueryParameter(e.getKey(), e.getValue());
    }
}
```

**To Reproduce**
A code sample or steps:
```kotlin
val options = SignInUIOptions.builder()
        .hostedUIOptions(HostedUIOptions.builder()
                .signInQueryParameters(mapOf("key" to "value"))
                .build())
        .build()
AWSMobileClient.getInstance().showSignIn(activity, options, callback)
```

**Which AWS service(s) are affected?**

**Expected behavior**
NullPointerException should not be happened

**Screenshots**

**Environment Information (please complete the following information):**
 - AWS Android SDK Version: 2.27.0
 - Device: Any
 - Android Version: Any
 - Specific to simulators: No

**Additional context**

raphkim:
Hi @monae ,

Thank you for catching this issue! I will patch it out as soon as possible and include it in our next release.

div5yesh:
The fix is released in v[2.28.0](https://github.com/aws-amplify/aws-sdk-android/releases/tag/release_v2.28.0)

PR #2564, opened 2021-08-04
fix(aws-android-sdk-mobile-client): fix NPE in hosted UI options

raphkim:
*Issue #, if available:*
#2560 
*Description of changes:*
Null-check was being performed on an irrelevant field, causing an NPE in certain situations. Now checks for `null` in correct field.


sdhuka (approved):
LGTM!

PR #2582, opened 2021-08-18
Set token expiration time when credentials provider is AWSMobileClient

sirknightj:
*Issue #, if available:*

*Description of changes:*
Updated the updateCredentials() method to take AWSMobileClient credentialsProvider into account. When the credentialsProvider is an instanceof AWSMobileClient, the returned expiration time is indefinite, meaning that this method does not get called in the future when the credentials expire and needs refreshing.

### DIFF
--- a/aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/mobileconnectors/kinesisvideo/auth/KinesisVideoCredentialsProviderImpl.java
+++ b/aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/mobileconnectors/kinesisvideo/auth/KinesisVideoCredentialsProviderImpl.java
@@ -28,8 +28,10 @@ import com.amazonaws.kinesisvideo.auth.KinesisVideoCredentials;
 import com.amazonaws.kinesisvideo.common.exception.KinesisVideoException;
 import com.amazonaws.kinesisvideo.common.logging.Log;
 import com.amazonaws.kinesisvideo.common.preconditions.Preconditions;
+import com.amazonaws.mobile.client.AWSMobileClient;
 
 import java.util.Date;
+import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


disa6302 (approved):
How many hours of run did we test with this change?

sirknightj:
> How many hours of run did we test with this change?

I left this running overnight, so about 8. Previously, it would stop after ~80 minutes, so we know that this is refreshing regularly.

disa6302:
Perfect. Thanks for confirming

aws-android-sdk-kinesisvideo/src/main/java/com/amazonaws/mobileconnectors/kinesisvideo/auth/KinesisVideoCredentialsProviderImpl.java
```diff
@@ -67,6 +69,15 @@ protected KinesisVideoCredentials updateCredentials() throws KinesisVideoExcepti
 
             expiration = cognitoCredentialsProvider.getSessionCredentialsExpiration();
             log.debug("Refreshed token expiration is %s", expiration);
+        } else if (credentialsProvider instanceof AWSMobileClient) {
+            AWSMobileClient awsMobileClient = (AWSMobileClient) credentialsProvider;
+            try {
```
poojamat, line 74: What happens to expiration time when credentials provider is none of these checked types?

sirknightj: The expiration date is then left at what it was previously: `KinesisVideoCredentials.CREDENTIALS_NEVER_EXPIRE` (`new Date(Long.MAX_VALUE)`), so this `updateCredentials()` method will basically never execute a second time.

poojamat (approved):

PR #2590, opened 2021-08-23
VersionInfoUtils null check for string replace

poojamat:
Adding a null check for string replace in VersionInfoUtils. This is causing crashes in production for kinesis.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws-android-sdk-core/src/main/java/com/amazonaws/util/VersionInfoUtils.java
+++ b/aws-android-sdk-core/src/main/java/com/amazonaws/util/VersionInfoUtils.java
@@ -131,6 +131,9 @@ public class VersionInfoUtils {
      * @return the input with spaces replaced by underscores
      */
     private static String replaceSpaces(final String input) {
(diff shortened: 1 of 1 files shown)


sdhuka (approved):
LGTM!

aws-android-sdk-core/src/main/java/com/amazonaws/util/VersionInfoUtils.java
```diff
@@ -131,6 +131,9 @@ static String userAgent() {
      * @return the input with spaces replaced by underscores
      */
     private static String replaceSpaces(final String input) {
-        return input.replace(' ', '_');
+        if (input != null){
+            return input.replace(' ', '_');
+        }
+        return input;
```
div5yesh, line 137: If we return null value here, `buffer.append(null)` will still throw an exception. We can return string "null" or empty string.

div5yesh (approved):

PR #2598, opened 2021-08-26
[Do not merge] transcribe Update models to latest

awsmobilesdk:
**[Do not merge] -- using only for testing a fix**


raphkim (changes requested):
Syntax errors in generated code

PR #2613, opened 2021-09-01
lambda: LambdaJsonBinder: Injectable Gson

yongjhih:
Allow inject Gson instance into `LambdaJsonBinder` constructor.


raphkim (commented):
Thank you for your contribution Andrew! The code changes look good to me, but may I know the use-case that requires this change to be made?

yongjhih:
> Thank you for your contribution Andrew! The code changes look good to me, but may I know the use-case that requires this change to be made?

Sure, the app has a singleton Gson, we want to reuse the gson instance for better performance, and also we have the customized typed serializers on it. Of course, we can just copy the current LambdaJsonBinder for that, but seems duplicated though.

raphkim (approved):
LGTM! 🚢

PR #2628, opened 2021-09-10
fix(cognitoidentityprovider):save lastAuthUser when using cached tokens

rjuliano:
*Issue #, if available:*

*Description of changes:*

If using multiple logins on the same device, the last auth user value was only getting cached if new tokens were acquired. So if you went through the following:

1 - Login with user1 (lastAuthUser = user1 and new tokens cached)
2 - Login with user2 (lastAuthUser = user2 and new tokens cached)
3 - Login with user1 again (lastAuthUser = user2 because new tokens were not issued since the ones from step 1 were still valid.

To correct this issue, I created a separate method that allows us to cache the lastAuthUser (without having to cache the tokens since that would be unnecessary).  This new method gets called from getCachedSession if the session being returned was from cache.

### DIFF
--- a/aws-android-sdk-cognitoidentityprovider/build.gradle
+++ b/aws-android-sdk-cognitoidentityprovider/build.gradle
@@ -8,13 +8,17 @@ android {
         minSdkVersion 9
         targetSdkVersion 29
     }
-
+    compileOptions {
+        sourceCompatibility 1.8
+        targetCompatibility 1.8
+    }
     testOptions.unitTests.includeAndroidResources = true
 }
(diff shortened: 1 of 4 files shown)


raphkim:
Can we get some tests to go with the code change?

rjuliano:
> Can we get some tests to go with the code change?

Yup...I'm working on those. Mocking stuff in the SDK is a little time consuming, so I wanted to get the fix out to see if anybody had any concerns.

raphkim (approved):

PR #2704, opened 2021-11-10
fix(Pinpoint): Add null check for endpoint response

changxu0306:
According to the [Pinpoint Documentation](https://docs.aws.amazon.com/pinpoint/latest/apireference/apps-application-id-events.html#apps-application-id-events-prop-eventsresponse-results), `Results` field of `EventsResponse` is not guaranteed to be present. The same assumption should also be made for the `.get(endpoint.getEndpointId())` call to avoid `NullPointerException`.

### DIFF
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
@@ -51,6 +51,7 @@ import com.amazonaws.services.pinpoint.model.Event;
 import com.amazonaws.services.pinpoint.model.EventItemResponse;
 import com.amazonaws.services.pinpoint.model.EventsBatch;
 import com.amazonaws.services.pinpoint.model.EventsRequest;
+import com.amazonaws.services.pinpoint.model.ItemResponse;
 import com.amazonaws.services.pinpoint.model.PublicEndpoint;
 import com.amazonaws.services.pinpoint.model.PutEventsRequest;
 import com.amazonaws.services.pinpoint.model.PutEventsResult;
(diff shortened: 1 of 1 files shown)


aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
```diff
@@ -436,16 +437,27 @@ private void submitEventsAndEndpoint(final JSONArray eventArray,
...
-        final EndpointItemResponse endpointItemResponse = resultResponse
+        final Map<String, ItemResponse> results = resultResponse
                 .getEventsResponse()
-                .getResults()
-                .get(endpoint.getEndpointId())
-                .getEndpointItemResponse();
-        if(202 == endpointItemResponse.getStatusCode()) {
-            log.info("EndpointProfile updated successfully.");
+                .getResults();
+        if (results.isEmpty()) {
```
sktimalsina, line 443: Looks like we are going to `if else` arrow here. Whenever a control does not need to continue after an `if` block, we can "short-circuit" the block by returning from there. This lets reader know that rest of the code is not executed and saves many tabs as bonus 😸

sktimalsina, line 443: Looks like we are going to `if else` arrow here. Whenever a control does not need to continue after an `if` block, we can "short-circuit" the block by returning from there. This lets reader know that rest of the code is not executed and saves many tabs as bonus 😸 
```suggestion
        if (results.isEmpty()) {
            log.error("PutEventsResult is empty!");
            return;
        } 
        
        if (endpoint.getEndpointId().isEmpty()) {
            log.error("EndpointId is missing.");
            return;
        } 
        
        final EndpointItemResponse endpointItemResponse = results
                .get(endpoint.getEndpointId())
                .getEndpointItemResponse();
        if (endpointItemResponse == null) {
            log.error("EndPointItemResponse is null!");
            return;
        }
        if (202 != endpointItemResponse.getStatusCode()) {
            log.error("AmazonServiceException occurred during endpoint update: " +
                endpointItemResponse.getMessage());
            return;
        } 
        log.info("EndpointProfile updated successfully.");
```

sktimalsina (changes requested):

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
changxu0306: Thanks @sktimalsina ! Short-circuit indeed makes the code cleaner 👍

sktimalsina (approved):

aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
```diff
@@ -436,12 +437,28 @@ private void submitEventsAndEndpoint(final JSONArray eventArray,
     }
 
     private void processEndpointResponse(EndpointProfile endpoint, PutEventsResult resultResponse) {
-        final EndpointItemResponse endpointItemResponse = resultResponse
+        final Map<String, ItemResponse> results = resultResponse
                 .getEventsResponse()
-                .getResults()
+                .getResults();
+        if (results == null || results.isEmpty()) {
+            log.error("PutEventsResult is empty!");
```
poojamat, line 444: Can we write a unit test for this?

poojamat (approved):

PR #2797, opened 2022-02-11
Fix 2634 : Adding the null guard when getTransferNetworkConnectionType is null

gpanshu:
*Issue #, if available:*
2634

*Description of changes:*

Adding a null guard for when getTransferNetworkConnectionType is null

### DIFF
--- a/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferRecord.java
+++ b/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferRecord.java
@@ -357,7 +357,7 @@ class TransferRecord {
      * @param connManager the android connectivity manager
      * @return true if the preferred network is available, false otherwise.
      */
-    private boolean checkPreferredNetworkAvailability(final TransferStatusUpdater updater,
+    protected boolean checkPreferredNetworkAvailability(final TransferStatusUpdater updater,
                                                       final ConnectivityManager connManager) {
 
         if (connManager == null) { 
--- a/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferRecord.java
+++ b/aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferRecord.java
@@ -364,9 +364,8 @@ class TransferRecord {
             // Unable to get the details of the network, we will start the transfer.
             return true;
         }
-            
-        if (transferUtilityOptions != null &&
-            !transferUtilityOptions.getTransferNetworkConnectionType().isConnected(connManager)) {
+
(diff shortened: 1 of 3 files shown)


aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferRecord.java
```diff
@@ -364,9 +364,8 @@ private boolean checkPreferredNetworkAvailability(final TransferStatusUpdater up
             // Unable to get the details of the network, we will start the transfer.
             return true;
         }
-            
-        if (transferUtilityOptions != null &&
-            !transferUtilityOptions.getTransferNetworkConnectionType().isConnected(connManager)) {
+
+        if (transferUtilityOptions != null && transferUtilityOptions.getTransferNetworkConnectionType() != null && !transferUtilityOptions.getTransferNetworkConnectionType().isConnected(connManager)) {
```
sktimalsina, line 368: Added null check will skip updating the state and return true for case when 
`transferUtilityOptions.getTransferNetworkConnectionType()` is null
is it expected?

poojamat: I think it would be a good idea to add a unit test on it as well.

sdhuka: Just returning without updating the state to WAITING_FOR_NETWORK might have unexpected outcome because TransferNetworkLossHandler queues all the tasks marked with WAITING_FOR_NETWORK when the network is available again. And also isConnected is deprecated and not reliable starting Android O. 

That said it is ok to skip updating the state incase of null to prevent the app from crashing.

sdhuka (approved):

aws-android-sdk-s3/src/main/java/com/amazonaws/mobileconnectors/s3/transferutility/TransferRecord.java
gpanshu: > Added null check will skip updating the state and return true for case when `transferUtilityOptions.getTransferNetworkConnectionType()` is null is it expected?

Yes as this will only return false when it is null and at that time the app will just crash

gpanshu: > Just returning without updating the state to WAITING_FOR_NETWORK might have unexpected outcome because TransferNetworkLossHandler queues all the tasks marked with WAITING_FOR_NETWORK when the network is available again. And also isConnected is deprecated and not reliable starting Android O.
> 
> That said it is ok to skip updating the state incase of null to prevent the app from crashing.

Yeah that code is still unchanged just added an extra null guard. Regarding the deprecated aspect since we support Jelly Bean and up we have to keep this code.

sktimalsina (approved):

poojamat (approved):

Issue #2370, opened 2021-02-25
Cognito User Pool password length error message should display user pool password length requirement


BillBunting:
**Which AWS Services is the feature request for?**
Cognito User Pools with AWSMobileClient DropIn UI 

**Is your feature request related to a problem? Please describe.**
Users have reported difficulty signing up for our app because they receive inaccurate or insufficient feedback from the DropIn UI sign up screen regarding password length requirements.

- Upon entering a 5 character password or less, the UI returns the error: "Sign up failed. Password should have 6 or more characters." However, the User Pool minimum password length is set to 8. Users are incorrectly told the password length is six. 
- Users that enter 6 or 7 character passwords, see a dialog "Sign Up failed. Password did not conform with policy. Password not long enough".

**Describe the solution you'd like**
- When a user enters a password less than the User Pool minimum password length, a dialog should inform them of the password length required by the user pool (not be hardcoded to 6)
- Users should not be told the minimum length is 6 (unless it actually is 6). (remove this validation message)

Here is the code to change:

```       
       if (password.length() < 6) {
            showError(getString(R.string.password_length_validation_failed));
            return;
        }
```
found at https://github.com/aws-amplify/aws-sdk-android/blob/cb623faaaab6d4dc8afe251762a135d02e9e3ffc/aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/SignUpActivity.java#L102

s16tom:
I've just come across the same thing and it is not only the wording that is inconsistent, which has caused a whole bunch of unexpected issues.

In the case of the 5 character password error, the presignup trigger doesn't run, any code in the trigger is not executed, a (slightly misleading) error is returned to the user and everyone is happy (ish; dodgy error message aside)!

In the case of the 6 or 7 character password error, the presignup trigger runs, gives the lambda code a user ID to execute with... then doesn't create the Cognito user and returns an error to the user. Everyone - especially the user that has a bunch of stuff provisioned for her and more so the admin that has to deal with their support call - is sad!

BillBunting:
Grooming my open issues: I removed the ability for new users of my app to create username/password accounts due to a few unresolved AWS UX experience issues (I received support requests/negative reviews from frustrated users dealing with the sign-up experience). Existing users may still sign-in with username/password, but all new users will sign-up with Facebook or Google.

PR #2800, opened 2022-02-14
Fix 2370 : Remove the password requirement

gpanshu:
*Issue #, if available:* #2370

*Description of changes:*
We will eventually be able to absorb the backend rules for password but for now I have aligned the code to iOS and removed the hardcoded min password requirement from the frontend. 

### DIFF
--- a/aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/SignUpActivity.java
+++ b/aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/SignUpActivity.java
@@ -99,11 +99,6 @@ public class SignUpActivity extends Activity {
             return;
         }
 
-        if (password.length() < 6) {
-            showError(getString(R.string.password_length_validation_failed));
-            return;
-        }
-
         final CognitoUserAttributes userAttributes = new CognitoUserAttributes();
         userAttributes.addAttribute(GIVEN_NAME, givenName);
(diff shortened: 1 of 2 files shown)


eeatonaws (commented):
Should we be removing [this section of code](https://github.com/aws-amplify/aws-sdk-android/blob/main/aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/CognitoUserPoolsSignInProvider.java#L411) as well that also checks the password length is at least 6 characters?

aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/SignUpActivity.java
```diff
@@ -99,11 +98,6 @@ public void signUp(final View view) {
             return;
         }
 
-        if (password.length() < 6) {
-            showError(getString(R.string.password_length_validation_failed));
-            return;
-        }
-
         final CognitoUserAttributes userAttributes = new CognitoUserAttributes();
```
div5yesh, line 101: Can we verify if appropriate error message will be shown if password length constraint is not met?

eeatonaws (commented):
The [string resource](https://github.com/aws-amplify/aws-sdk-android/blob/main/aws-android-sdk-auth-userpools/src/main/res/values/strings.xml#L45) with the password length message can be removed.

PR #2961, opened 2022-07-27
Fix: Prevent startForeground crash and TransferService stopping while transfers ongoing

tylerjroach:
*Issue #, if available:*

*Description of changes:*
Seeing integration test issues that `Context.startForegroundService() did not then call Service.startForeground()`. The interesting thing about this is that the foreground service has already been started and made an initial call to `startForeground()`. What is not clear is if each call to `startForegroundService` needs to be followed with `startForeground()`. We have not observed any manual crashes due to this.

The check on `hasNotificationShown` was done to prevent continually overwriting the notification, but overwriting isn't a bad thing.


sdhuka (approved):

gpanshu (approved):

tylerjroach:
On testing, this will cause sound/vibrate each time notification is replaced, which is not optimal. Need to consider changing notification channel for silent notifications.

sdhuka (approved):

gpanshu (approved):

PR #2971, opened 2022-08-05
Fix Javadocs jar file publishing using buildspec file.

### LINKED ISSUE #2929: JavaDocs no longer published as of 2.18.x.
After the change to use Gradle Maven publishing plugin (as of 2.18.0), JavaDocs are no longer being published.

This situation is also true of the Amplify Android project, which uses a very similar publishing script.

https://github.com/aws-amplify/amplify-android/issues/651

This is ___not___ true of the AppSync SDK, though.

https://repo1.maven.org/maven2/com/amazonaws/aws-android-sdk-appsync-gradle-plugin/3.1.1/

Likely we need only to make some minor changes in `publishing.gradle`, to include artifact specfications similar to this:

https://github.com/awslabs/aws-mobile-appsync-sdk-android/blob/main/gradle-mvn-push.gradle#L228-L237

tjleing:
Fixes #2929.  Once merged, we need to change the build process to use the `maven-release-publisher.yml` file as a buildspec.  Unless the team feels otherwise, I will attempt a dry run release (into Sonatype but not released to Maven) to test these changes and make sure they work.  I would also like to shadow the next oncall's release process.

There are hundreds of Javadocs errors visible when running the command due to more stringent HTML tag requirements as well as some things just not having been checked.  We can fast-follow on fixing those issues -- for now we have `failOnError false` to skip over these errors.  I have verified that the errors do not prevent the Javadocs from being generated/used, although of course I have not checked every single error.


publishing.gradle
```diff
@@ -156,10 +156,17 @@ afterEvaluate { project ->
         }
 
         task androidJavadocs(type: Javadoc) {
+            failOnError false
```
gpanshu, line 159: What happens when it fails then?

div5yesh (approved):

publishing.gradle
tjleing: The Javadocs still generate, I have a bit more detail in the PR description.  Many of the issues are due to an increase in the strictness of the HTML parser, when browsers support the features (auto-closed <br/> tags, nested tags, <code> tags at all, etc.).  To do a full audit of the issues would take much more time given their quantity.

tylerjroach (approved):

Issue #2255, opened 2020-11-23
Error callback in Hosted UI function not getting proper error message from Lambda function


shishir87:
**AWS Cognoto Hosted UI**

**Is your feature request related to a problem? Please describe.**
Hi, 
I am developing an Android app that has login with Google and Facebook options. To achieve this I am using below line of code using Kotlin language.

```
val hostedUi = HostedUIOptions.builder()
    .scopes("openid", "email")
    .identityProvider("Facebook") // here it can be "Google" for login with google
    .build()
val signInUIOptions = SignInUIOptions.builder()
    .hostedUIOptions(hostedUi)
    .build()
AWSMobileClient.getInstance()
    .showSignIn(requireActivity(), signInUIOptions, object : Callback<UserStateDetails> {
        override fun onResult(details: UserStateDetails) {
        }
        override fun onError(e: Exception) {
            Log.e(TAG, "onError: ", e)    //   [1] I should get the lambda exception here 
        }
    }
)
```
Our backend developer has written a lambda trigger to check whether the same email Id is already present in database when user sign up and if it does, then it throws an error with error_description = "PreSign up failed with error An account with the given email already exists".

Now my problem is, I am not getting this "error_description" message in onError callback method. (refer [1] in kotlin code given above).
Instead, I get AuthServiceException with message "invalid_request". I want the exact same "error_description" that lambda is throwing.
Please help to solve my problem.

Thanks,
Shishir Pednekar

eeatonaws:
A fix for the issue has been released in version 2.53.0 of the AWS SDK for Android.

PR #3002, opened 2022-09-09
fix(cognitoauth): forward lambda trigger exception for social sign in

eeatonaws:
*Issue #, if available:* #2255, [1649](https://github.com/aws-amplify/amplify-android/issues/1649)

*Description of changes:* When an exception is caused by a lambda trigger for social sign in, the resulting URI contains an "error" parameter equal to "invalid_request" and an "error_description" parameter that contains the actual error message from the exception thrown in the lambda. The message for the exception returned to the onError for social sign in is currently "invalid_request" (the value of the "error" parameter), so this PR changes the message to be the "error_description" instead when the error is "invalid_request".


sktimalsina (approved):
🚢 🚀

sdhuka (approved):

Issue #3018, opened 2022-09-23
Amazon S3 - The authorization header is malformed; the region 'vpce' is wrong; expecting 'eu-central-1'


heisenbergye:
**Describe the bug**
The authorization header is malformed; the region 'vpce' is wrong; expecting 'eu-central-1'

**To Reproduce**
A code sample or steps:
![image](https://user-images.githubusercontent.com/25834939/191966407-830eed26-7f61-41ff-ba0d-91a4abc36cbd.png)

when creating a new progress listener for transfer, it got an ERROR as the following:
```
Received error response: com.amazonaws.services.s3.model.AmazonS3Exception: The authorization header is malformed; the region 'vpce' is wrong; expecting 'eu-central-1' (Service: null; Status Code: 400; Error Code: AuthorizationHeaderMalformed; Request ID: DN55*******2CRZ4), S3 Extended Request ID: RN8eBGO2ZCJGGy*******KtqO/1DwofYb6Bim2giwzG*******EiHZBliYfeNcU9*******OPblJJFQ=
```

maybe it is a bug similar as :
https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/troubleshooting_s3a.html
![image](https://user-images.githubusercontent.com/25834939/191967319-de20261a-85cc-485a-ba7a-3ed58150f8e5.png)

**Which AWS service(s) are affected?**
when upload picture by S3 interface endpoint in VPC, the transfer cannot be created.

**Expected behavior**
Please help to fix it that parse region code and produce url correctly. 

**Environment Information (please complete the following information):**
 - AWS Android SDK Version: 2.35.0, 2.53.0

eeatonaws:
Thank you for reporting this issue. The AmazonS3Client is indeed parsing endpoints in VPC such that the region is "vpce". I am working on a fix for this issue.

eeatonaws:
A fix for the issue has been released in AWS SDK for Android 2.54.0.

EddyChina:
any update on AWS SDK for Java? I got same exception with flink checkpoint.

eeatonaws:
Please create an issue in the [AWS SDK for Java GitHub repository](https://github.com/aws/aws-sdk-java/issues/new/choose).

PR #3024, opened 2022-09-28
fix(aws-android-sdk-core): modify region parsing for endpoints in vpc

eeatonaws:
*Issue #, if available:* #3018

*Description of changes:* When parsing an endpoint in VPC to retrieve the region, the existing region parsing logic would determine the region in an endpoint such as "bucket.vpce-1234.s3.us-west-2.vpce.amazonaws.com" is "vpce" but it should be "us-west-2". This PR modifies the region parsing logic to check the case when the endpoint is in VPC.

To test these changes, I added a new test for parsing an endpoint in VPC and verified that all existing tests in AwsHostNameUtilsTest.java still pass.


div5yesh (approved):
LGTM

tylerjroach (approved):

PR #3060, opened 2022-11-04
Correctly pass Pinpoint ID (or null if it's not there) to Cognito User Pool

tjleing:
**TESTED!**

Before:
- no Pinpoint app id: creates SharedPreferences file with 515... name and random endpoint id <rEPID>, creates SharedPreferences file with <rEPID>515... name and random endpoint id <rrEPID>
- with Pinpoint app id: creates or SharedPreferences file with <appID>515... name and endpoint id <rEPID>, creates SharedPreferences file with <rEPID>515... name and random endpoint id <rrEPID>

After:
- no Pinpoint app id: no SharedPreferences file, no Pinpoint endpoint passed with Cognito calls
- with Pinpoint app id: creates or reads SharedPreferences file with <appID>515... name and endpoint id <rEPID>, uses that endpoint with Cognito calls

[Docs update](https://github.com/aws-amplify/docs/pull/4780)


tylerjroach (approved):
We have agreed upon this change because unfortunately Pinpoint SDK is sending a different endpointId than AWSMobileClient. If this were just a matter of constructing/creating the endpointId in a weird way, I would advocate to try and migrate that value correctly, but the issue here is that we already have an endpointId properly in use by the Pinpoint SDK, so there is no resolution other than beginning to use the correct endpointId.

sdhuka (approved):

PR #3191, opened 2023-02-03
chore(analytics): Simplify error code checking, uniformly across platforms

tjleing:
Flutter team asked Pinpoint about error codes and discovered that any event with a status code of 5xx can be retried. Checking specifically for exception names may be deprecated if they decide to add additional exception types, but they will maintain that 5xx error codes can be retried going forward. Matching PR on Amplify v2 [here](https://github.com/aws-amplify/amplify-android/pull/2274).

### DIFF
--- a/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
+++ b/aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
@@ -438,22 +438,22 @@ public class EventRecorder {
         } catch (final AmazonServiceException amazonServiceException) {
             // This is service level exception, we also have item level exception.
             log.error("AmazonServiceException occurred during send of put event ", amazonServiceException);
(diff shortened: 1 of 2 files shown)


aws-android-sdk-pinpoint/src/main/java/com/amazonaws/mobileconnectors/pinpoint/internal/event/EventRecorder.java
```diff
@@ -548,13 +548,8 @@ private void processEventsResponse(final JSONArray eventArray,
...
     }
 
-    private boolean isRetryable(String responseCode) {
-        if (responseCode.equalsIgnoreCase("ValidationException") ||
-            responseCode.equalsIgnoreCase("SerializationException") ||
-            responseCode.equalsIgnoreCase("BadRequestException")) {
-            return false;
-        }
-        return true;
+    private boolean isRetryable(int httpCode) {
```
sdhuka, line 551: could we add a unit test to cover this change?

mattcreaser (approved):

sdhuka (approved):

PR #3588, opened 2024-06-06
fix(aws-android-sdk-auth-userpools): Check actual password requirements in drop-in UI

mattcreaser:
*Issue #, if available:* #2370

*Description of changes:*
Don't hardcode the minimum password length to six. Instead, check the `passwordProtectionSettings` defined in `awsconfiguration.json` and use those for client-side validation.

Screenshot_20240605_153943

### DIFF
--- a/aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/SignUpActivity.java
+++ b/aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/SignUpActivity.java
@@ -101,7 +101,7 @@ public class SignUpActivity extends Activity {
         Log.d(LOG_TAG, "phone = " + phone);
 
 
-        final Integer minimumPasswordLength = getMinimumPasswordLength();
+        final Integer minimumPasswordLength = CognitoUserPoolsSignInProvider.getMinimumPasswordLength(configuration);
 
         if (username.isEmpty()) {
             showError(getString(R.string.sign_up_username_missing));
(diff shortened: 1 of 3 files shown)


lawmicha (approved):
LGTM

aws-android-sdk-auth-userpools/src/main/java/com/amazonaws/mobile/auth/userpools/SignUpActivity.java
```diff
@@ -94,13 +100,16 @@ public void signUp(final View view) {
...
+        final Integer minimumPasswordLength = CognitoUserPoolsSignInProvider.getMinimumPasswordLength(configuration);
+
         if (username.isEmpty()) {
             showError(getString(R.string.sign_up_username_missing));
             return;
         }
 
-        if (password.length() < 6) {
-            showError(getString(R.string.password_length_validation_failed));
+        if (minimumPasswordLength != null && password.length() < minimumPasswordLength) {
```
tylerjroach, line 111: when would minimum be null? Should we fall back to 6 as a safety? Is lower than 6 ever allowed?

mattcreaser: If the password requirements are not specified in `awsconfiguration` this can be null.

6 is the minimum in Cognito, but if we don't know what the customer has set as their minimum then we just don't want to validate client-side, since any value we choose is likely to be wrong. The original customer report complains that validating with the wrong length leads to a confusing end user experience.

So if we don't know what the minimum is we'll just let the server side validate the password and display whatever error message they return if it's too short (this part is existing functionality).

tylerjroach (approved):

Issue #3560, opened 2024-04-08
Consumer proguard rules are too broad


alipov:
**Describe the bug**
SDK's [consumer proguard rules](https://github.com/aws-amplify/aws-sdk-android/blob/main/aws-android-sdk-core/consumer-proguard-rules.pro#L6) include the following rule that affects all enum classes in hosting application (not just the ones belonging to this SDK):  
```
# Enums are not obfuscated correctly in combination with Gson
-keepclassmembers enum * { *; }
```
There are enum classes in my application whose members I want to be obfuscated, but the above rule prevents that.

**To Reproduce**
Create an app with an enum class, consume it somewhere. Build a release version with minification/obfuscation enabled.

**Which AWS service(s) are affected?**
None, but it affects the hosting app.

**Expected behavior**
The SDK doesn't affect the obfuscation of host application's enum members.  
For example, consider to reduce rule's scope only to classes in `com.amazonaws` or `com.amazon` packages:  
```
-keepclassmembers enum com.amazonaws.**,com.amazon.** { *; }
```

**Environment Information (please complete the following information):**
 - AWS Android SDK Version: latest
 - Device: any
 - Android Version: any
 - Specific to simulators: no

joon-won:
Hi @alipov, our team will take a look into the issue

PR #3619, opened 2024-07-26
chore: proguard rule for keepclassmembers enum

lawmicha:
*Issue #, if available:*

*Description of changes:*
https://github.com/aws-amplify/aws-sdk-android/issues/3560


mattcreaser:
This looks good, but we need to verify that there are no enums outside of these packages that are used in data serialized by Gson 😬

PR #3677, opened 2025-09-22
fix: Update native libraries with 16KB page alignment support

YPandas:
Update native libraries with 16KB page alignment support

## Background
This change addresses compatibility issues with newer Android devices that use 16KB memory pages.

## Changes
- Replace .so files for arm64-v8a, armeabi-v7a, x86, and x86_64 with 16KB-aligned versions
- Add ANDROID_SUPPORT_FLEXIBLE_PAGE_SIZES compiler flag for runtime compatibility
- Configure linker flags: max-page-size=16384, common-page-size=4096

## Testing
- Verified libraries load correctly on devices with both 4KB and 16KB page sizes
- Used native build v3.1.1 to build the .so files and tested streaming functionality on Android emulator - confirmed working correctly


tylerjroach (approved):

tylerjroach (changes requested):
Looks like we still have an armeabi folder with an old .so file. Can you update that as well?.

tylerjroach (approved):

tylerjroach:
That should be good especially given min sdk version for KinesisVideo is 24.

sirknightj (approved):

rashworld-max (approved):

